PR #3, opened 2018-06-08
update parsing of SMTlib to version 2.6 for constructs required by SL…

mihasighi:
…-COMP


Gbury:
Sorry for the delay. This looks good.
Out of curiosity, does this allow to parse all of SMTLIB v2.6 ? And if not, how hard would it be to parse all the specification of SMTLIB v2.6 ?

Gbury:
Oh, actually, I just read the SMTLIB v2.6 specification, and if I understand correctly `declare-heap` is an extension, and not part of the core SMTLIB ?

src/languages/smtlib/lexSmtlib.mll
```diff
@@ -42,35 +42,53 @@ rule token = parse
...
-  | "set-option"        { SET_OPTION }
-  | "set-info"          { SET_INFO }
-  | "declare-sort"      { DECLARE_SORT }
-  | "define-sort"       { DEFINE_SORT }
-  | "declare-fun"       { DECLARE_FUN }
-  | "define-fun"        { DEFINE_FUN }
-  | "push"              { PUSH }
-  | "pop"               { POP }
+
+  | "not"               { NOT }
```
Gbury: Why is `not` singled out ? As far as I can tell, the SMTLIB specification does not treat `not` specifically (from a syntax point of view) and I'd like to keep it that way if possible (i.e. keep the lexer/parser as close to the formal syntax definition as possible).

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -114,11 +128,30 @@ sorted_var:
...
+    { let f = I.(mk term s) in
+      f, (args, ty) }
+
 term:
   | c=spec_constant
     { c I.term }
   | s=qual_identifier
     { let loc = L.mk_pos $startpos $endpos in T.apply ~loc s [] }
+  | OPEN NOT t=term CLOSE
+    { let loc = L.mk_pos $startpos $endpos in T.apply ~loc (T.const ~loc I.(mk term "not")) [t] }
```
Gbury: Same as in the lexer, why do you treat `not` specially ?

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -137,60 +170,109 @@ command_option:
...
+  | OPEN DECLARE_CONST s=SYMBOL ty=sort CLOSE
+    { let id = I.(mk term s) in
+      let loc = L.mk_pos $startpos $endpos in
+      S.fun_decl ~loc id [] ty }
 
-  | OPEN GET_INFO i=KEYWORD CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.get_info ~loc i }
-  | OPEN SET_INFO c=command_option CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.set_info ~loc c }
+  | OPEN DECLARE_DATATYPE s=SYMBOL OPEN clist=datatype_dec CLOSE CLOSE
```
Gbury: The parentheses around the datatype_dec seems superfluous (as datatype_dec already starts and ends wiht parentheses), and not part of the SMTLIB syntax specification.

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -137,60 +170,109 @@ command_option:
...
+  | OPEN DEFINE_FUN_REC s=SYMBOL OPEN args=sorted_var* CLOSE ret=sort body=term CLOSE
+    { let id = I.(mk term s) in
+      let loc = L.mk_pos $startpos $endpos in
+      let rty = T.annot ~loc ret [T.const ~loc I.(mk attr "rec")] in
+      S.fun_def ~loc id args rty body }
+
+  | OPEN DEFINE_FUNS_REC OPEN fsl=fun_dec+ CLOSE OPEN fbl=term+ CLOSE CLOSE
+    { let loc = L.mk_pos $startpos $endpos in
+      let fd = List.map2 
+          (fun fs fb -> S.fun_def ~loc (fst fs) (fst (snd fs)) (T.annot ~loc (snd (snd fs)) [T.const ~loc I.(mk attr "rec")])  fb)
```
Gbury: That line is quite long, maybe split it / introduce some let-bindings to make it more readable ?

src/standard/statement.ml
```diff
@@ -243,6 +247,10 @@ let assumption ?loc l =
 let check_sat = prove
 let assert_ ?loc t = antecedent ?loc t
 
+let dtype_decl ?loc id pl cstrs =
```
Gbury: Mabe the `dtype_dec` function could take a list of terms (rather than a list of ids) as argument `pl` (this is made so that correct location can be read from the `vars` field) ?

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -137,60 +170,109 @@ command_option:
...
+      S.fun_decl ~loc id [] ty }
 
-  | OPEN GET_INFO i=KEYWORD CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.get_info ~loc i }
-  | OPEN SET_INFO c=command_option CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.set_info ~loc c }
+  | OPEN DECLARE_DATATYPE s=SYMBOL OPEN clist=datatype_dec CLOSE CLOSE
+    { let id = I.(mk sort s) in
+      let loc = L.mk_pos $startpos $endpos in
+      S.dtype_decl ~loc id [] (snd clist) }
```
Gbury: Maybe assert here that `(fst clist) = []` ? (probably splitting `clist` with a let first), to make it clear that no information is discarded.

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -137,60 +170,109 @@ command_option:
...
 
-  | OPEN GET_OPTION k=KEYWORD CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.get_option ~loc k }
-  | OPEN SET_OPTION c=command_option CLOSE
-    { let loc = L.mk_pos $startpos $endpos in S.set_option ~loc c }
+  | OPEN DECLARE_DATATYPES OPEN sl=sort_dec+ CLOSE OPEN lcl=datatype_dec+ CLOSE CLOSE
+    { let loc = L.mk_pos $startpos $endpos in
+      let dl = List.map2 (fun sd cl -> S.dtype_decl ~loc (fst sd) (fst cl) (snd cl)) sl lcl
```
Gbury: This function is quite hard to read and understand.
It would be better to plit it over multiple lines, and use let-bindings instead of `fst` or `snd`, in order to be more explicit (i.e. give a legible name to the intermediary variables), and be a tinly little bit more efficient (let's call it a personnal preference, ^^)

src/languages/smtlib/parseSmtlib.mly
```diff
@@ -137,60 +170,109 @@ command_option:
...
+    
+  | OPEN DEFINE_FUN s=SYMBOL OPEN args=sorted_var* CLOSE ret=sort body=term CLOSE
+    { let id = I.(mk term s) in
+      let loc = L.mk_pos $startpos $endpos in
+      S.fun_def ~loc id args ret body }
+
+  | OPEN DEFINE_FUN_REC s=SYMBOL OPEN args=sorted_var* CLOSE ret=sort body=term CLOSE
+    { let id = I.(mk term s) in
+      let loc = L.mk_pos $startpos $endpos in
+      let rty = T.annot ~loc ret [T.const ~loc I.(mk attr "rec")] in
```
Gbury: For this line, it would be better to use the location for `ret` rather than the location of the whole command. This is easily done usgin `$startpos(ret)` and `$endpos(ret)`

src/languages/smtlib/lexSmtlib.mll
```diff
@@ -42,35 +42,53 @@ rule token = parse
...
+
   | "assert"            { ASSERT }
+  | "check-sat-assuming"        { raise Error }
```
Gbury: Could you add the corresponding token (and associated syntax rules), rather thna raising Error ? (Not absolutely necessary as I can do it after merging the PR)

src/languages/smtlib/lexSmtlib.mll
```diff
@@ -42,35 +42,53 @@ rule token = parse
...
   | "get-unsat-core"    { GET_UNSAT_CORE }
   | "get-value"         { GET_VALUE }
-  | "get-assignment"    { GET_ASSIGNMENT }
-  | "get-option"        { GET_OPTION }
-  | "get-info"          { GET_INFO }
-  | "exit"              { EXIT }
+  | "pop"               { POP }
+  | "push"              { PUSH }
+  | "reset-assertions"  { raise Error }
+  | "reset"             { raise Error }
```
Gbury: Same as for `check-sat-assuming`.

src/languages/smtlib/lexSmtlib.mll
```diff
@@ -42,35 +42,53 @@ rule token = parse
...
+  | "define-sort"       { DEFINE_SORT }
+  | "echo"              { raise Error }
+  | "exit"              { EXIT }
   | "get-assertions"    { GET_ASSERTIONS }
+  | "get-assignment"    { GET_ASSIGNMENT }
+  | "get-info"          { GET_INFO }
+  | "get-option"        { GET_OPTION }
+  | "get-model"         { GET_MODEL }
   | "get-proof"         { GET_PROOF }
+  | "get-unsat-assumptions"    { raise Error }
```
Gbury: Same as for `check-sat-assuming`.

src/languages/smtlib/lexSmtlib.mll
```diff
@@ -42,35 +42,53 @@ rule token = parse
...
+  | "declare-datatypes" { DECLARE_DATATYPES }
+  | "declare-datatype"  { DECLARE_DATATYPE }
+  | "declare-fun"       { DECLARE_FUN }
+  | "declare-heap"      { DECLARE_HEAP }
+  | "declare-sort"      { DECLARE_SORT }
+  | "define-fun-rec"    { DEFINE_FUN_REC }
+  | "define-funs-rec"   { DEFINE_FUNS_REC }
+  | "define-fun"        { DEFINE_FUN }
+  | "define-sort"       { DEFINE_SORT }
+  | "echo"              { raise Error }
```
Gbury: Same as for `check-sat-assuming`.

mihasighi:
Dear Guillaume,

The current parsing parse SMTLIB v2.6 except for the constructs that are 
not
allowed by your AST. It will not be very complex to add them.

Indeed, 'declare-heap' is an extension because Separation Logic is a
2nd order logic and needs a command to type the 2nd order var
which is the heap.

Best,
Mihaela

Gbury:
Feel free to disregard the comments about constructs not allowed in the AST if you do not want to add them yourself, I can do it after merging the PR.

Gbury:
Also, it would be interesting to document somewhere in the doc that the parser actually handles an extension of SMTLIB that includes the `declare-heap` command, do you have any page/pdf documenting the extension that the doc could link to ?

mihasighi:
Yes, the "official" reference is on GitHub also:

https://github.com/sl-comp/SL-COMP18/tree/master/input/Docs/smtlib-sl.pdf

mihasighi:
Hi Guillaume,

I'm sorry being in late with my comments on your comments : I'm running 
our competition this week-end, so much things to to. I hope to have some 
time next week.

best,
Mihaela

Gbury:
I'm actually working on fully upgrading the parser to smtlib v2.6 (and cleaning up a few things along the way), so maybe wait a litte bit before you start working on this PR again, ^^

src/languages/smtlib/lexSmtlib.mll
mihasighi: It is fine to me. This shall be dealt at the lecture of the theory (e.g., 'not' is in the core theory) for the type checker.

src/languages/smtlib/parseSmtlib.mly
mihasighi: Fine for me.

src/languages/smtlib/parseSmtlib.mly
mihasighi: You're right.

Gbury:
So, I just pushed in c19e70731ddb82f1a4c571cf4ef1795217da7dd8 an update to parse all of SMTlib v2.6.

If I understand correctly, it is not enough for your needs, as you need the additional `declare-heap` command, right ?

mihasighi:
> So, I just pushed in c19e707 [1] an update to parse all of SMTlib
> v2.6.

Thank you, I'll try it shortly.

> 
> If I understand correctly, it is not enough for your needs, as you
> need the additional declare-heap command, right ?
> 

You're right! This command is the only addition to the SMT-lib format.

Gbury:
Closing this one for now, as it's a bit old with no activity, and dolmen now correctly parses and types the smtlib v2.6.

If you want/need dolmen to support smtlib extensions, don't hesitate to ask (we've already added psmt2, so it's not a problem to add new smtlib extensions, ^^).

PR #4, opened 2018-07-13
Typechecking: first try

Gbury:
First attempt at integrating a typecheker in Dolmen.

The interface of the functor involve some quite long constraints, but I didn't see any good solution to avoid these.

A few design decision actually came from the fact that I actually extracted the typechecker from Archsat. They are not strictly mandatory, so here is a (non-exhaustive) list of things to criticize:
- tags are used to represent attributes, but that requires some involvements in all modules (types, terms, and formulas)
- requiring substitutions may seem a bit weird: they are use because of smtlib's aliases. Instead of textually substituting them, and then re-perform type checking, the body of the alias is typechecked, and then the typed terms are substitute at each occurence of an alias. It seemed cleaner to do it this way, but it requires to knwo about substitutions in the typechecker
- also about substitutions: to correctly substitute in a polymorphic term, you need to substitute both type variables and term variables, currently split into two substitution types (which adds some type constraints in the functor signature)
- the typechecker can provide typo suggestions for symbols not found, but this requires an additional dependency: `spelll`, is it worth it ?
- Errors in typechecking are all represented as the same exception with a string message. Given the structure of the typechecker, to provide helpful messages, this requires some structures (but not all) to provide some printing functions. This non-symmetry bothers me a bit. Would it be better to either require print functions for all structures, or use a more structures error representation to avoid having to print structures at all ?
- The global state of the typechecker is currently hidden from the user, which would be a problem for handling some smtlib problems where iirc, type definitions may be backtracked (because of pushs/pops)

Still on the todo list:
- [ ] add the builtins definitions for tptp, smtlib (including arithmetic ?)
- [ ] add a lot more documentation
- [ ] provide a std implementation of typed terms ?
- [ ] integrate it to the binary to also check typing
- [ ] CI tests for typechecking
- [ ] Add builtins for all smtlib theories ?

cc @bobot @c-cube @mihasighi


mihasighi:
My request for the type checking is to do it after reading the theory files. 
This feature is very useful for Separation Logic theory where the signature is not very usual.
The idea is to read into a table the symbols (sorts and functions) introduced by the theories of the logic declared by 'set-logic'. The type checker  should work in the presence of built-ins and the new signature.

Gbury:
> My request for the type checking is to do it after reading the theory files.

Okay, to do that is actually easy (the typechecker is easily extensible using the 'builtins' field of the local environment). It should be fairly easy to also give functors to create an appropriate `builtin` function.

However, I have a question first: these symbols defined in a specific theory file for separation logic, are they supposed to be particularly interpreted by provers (in the general case), or are the provers supposed to read this theory file in order to read and use (in a generic first-order way) the axioms written in it ?

For instance, while arithmetic symbols are also defined in a theory file by SMTLIB, any prover that seriously aims at handling arithmetic symbols will have a builtin theory that uses a specific representation and a hardcoded interpretation of the symbols defined by the arithmetic theory (i.e. have some decision procedure implemented in the prover, rather than reading and using a first-order axiomatisation of arithmetic). Thus in order to handle arithmetic terms with this typechecker, the functor for the `builtin` function should take as argument enough function to build the specialized representation of arithmetic terms.

My question is then, should it be the same for separation logic (where the user has to provide some functions to create the separation logic specific terms), or should the `builtin` function simply read the theory file and add the symbols to the environment (in which case, after typechecking, they would be treated as any arbitrary symbol defined by the input problem).

Gbury:
Also, as a side-note, how frequent is this theory file supposed to change ? I fear that having to read, parse and treat this theory file for every input file when typechecking could be quite a waste of time (and could be easily avoided by hardcoding the defined symbols in some provided `builtin` function).

mihasighi:
The builtin symbols of SL are defined in a specific theory file and are interpreted by the provers, no axiomatisation is given. The user may define other predicate (function with return type Bool), datatype or sort symbols, in addition to built-ins.

The theory file has changed this year and I hope that it is now stabilised. But there is no consensus yet on what is the best theory for SL in SMT-lib.

Gbury:
Just update this PR with the following:
- removed the need for substitutions in favor of a `letin` term constructor
- stopped distinguishing terms and formulas: since let-bindings can bind terms in formulas and vice-versa, maintaining the distinction was becoming a burden.

PR #13, opened 2018-11-06
Replace ocamlbuild with dune: First try

### LINKED ISSUE #12: Switch to dune ?
Providing printers for terms needs to correctly handle unicode[1] in the general case. Thus it adds uutf and uucp as dependencies, however people may not need or want those dependencies in a parser library[2].

The simple answer is to provide another package (kind of a sub-package) that provides the export functionalities, and which can depend on uutf and uucp while the main package does not need them.

However, correctly handling multiple packages is quite a pain to deal with when using ocamlbuild and manual makefiles for installation, so switching to dune would probably ease that a lot (once the switch done), so here it goes: dolmen should probably switch to dune.

If anyone is motivated to do the switch, please do so and create a PR ! Else I'll have a look sometime soon.

cc @anmaped

[1] : handling unicode primarily means detecting unicode characters in order to know what to do with them in languages that do not support unicode. This require a unicode-aware handling of identifiers strings, if only to replace unicode characters by appropriate ASCII characters.

[2]: though if dolmen one day parses unicode-aware languages (such as Coq ?), a unicode-ware lexer might be needed...

Gbury:
Using the dune file suggested in #12. Should close #12 once it is merged.

However, it does not compile yet...

Current error:

```
Error: No rule found for src/lexDimacs.mll
Error: No rule found for src/parseDimacs.mly
```

Maybe following dune's conventions and having one dune file (and one library) per directory would be the way to go. Dolmen already has something close to that as far as the hierarchy of modules is concerned.


anmaped:
```
(rule
  (targets lexiCNF.ml)
  (deps    src/languages/icnf/lexiCNF.mll)
  (action  (chdir %{workspace_root}
            (run %{bin:ocamllex} -q -o %{targets} %{deps})))
)

(rule
  (targets lexZf.ml)
  (deps    src/languages/zf/lexZf.mll)
  (action  (chdir %{workspace_root}
            (run %{bin:ocamllex} -q -o %{targets} %{deps})))
)

(rule
  (targets lexTptp.ml)
  (deps    src/languages/tptp/lexTptp.mll)
  (action  (chdir %{workspace_root}
            (run %{bin:ocamllex} -q -o %{targets} %{deps})))
)

(rule
  (targets lexSmtlib.ml)
  (deps    src/languages/smtlib/lexSmtlib.mll)
  (action  (chdir %{workspace_root}
            (run %{bin:ocamllex} -q -o %{targets} %{deps})))
)

(rule
  (targets lexDimacs.ml)
  (deps    src/languages/dimacs/lexDimacs.mll)
  (action  (chdir %{workspace_root}
            (run %{bin:ocamllex} -q -o %{targets} %{deps})))
)

(rule
  (targets parseiCNF.ml)
  (deps    src/languages/icnf/parseiCNF.mly  src/languages/icnf/tokens_iCNF.mly)
  (action  (chdir %{workspace_root}
            (run %{bin:menhir} --explain %{deps} --base parseiCNF)))
)

(rule
  (targets parseZf.ml)
  (deps    src/languages/zf/parseZf.mly  src/languages/zf/tokens_zf.mly)
  (action  (chdir %{workspace_root}
            (run %{bin:menhir} --explain %{deps} --base parseZf)))
)

(rule
  (targets parseTptp.ml)
  (deps    src/languages/tptp/parseTptp.mly  src/languages/tptp/tokens_tptp.mly)
  (action  (chdir %{workspace_root}
            (run %{bin:menhir} --explain %{deps} --base parseTptp)))
)

(rule
  (targets parseSmtlib.ml)
  (deps    src/languages/smtlib/parseSmtlib.mly  src/languages/smtlib/tokens_smtlib.mly)
  (action  (chdir %{workspace_root}
            (run %{bin:menhir} --explain %{deps} --base parseSmtlib)))
)

(rule
  (targets parseDimacs.ml)
  (deps    src/languages/dimacs/parseDimacs.mly  src/languages/dimacs/tokens_dimacs.mly)
  (action  (chdir %{workspace_root}
            (run %{bin:menhir} --explain %{deps} --base parseDimacs)))
)

(library
  (name            dolmen)
  (public_name     dolmen)
  (modules

; Standard implementation
Id
Term
Normalize
Statement
ParseLocation

; Interfaces
Lex_intf
Parse_intf
Location_intf

Id_intf
Term_intf
Stmt_intf
Language_intf


; Dimacs language
Dimacs
LexDimacs
ParseDimacs
Ast_dimacs

; iCNF language
ICNF
LexiCNF
ParseiCNF
Ast_iCNF

; Smtlib language
Smtlib
LexSmtlib
ParseSmtlib
Ast_smtlib

; Zf language
Zf
LexZf
ParseZf
Ast_zf

; Tptp language
Tptp
LexTptp
ParseTptp
Ast_tptp

; Languages classes
Logic

; Helper modules
Misc
Transformer

  )
  (libraries       MenhirLib)
)

(include_subdirs unqualified)
```

anmaped:
If I unroll the rules manually It works `(include_subdirs unqualified)` is not working for menhir!

anmaped:
But we have several issues like that.
```
File "src/standard/id.ml", line 21, characters 11-13:
Warning 27: unused variable ns.
File "src/standard/id.ml", line 24, characters 16-18:
Warning 27: unused variable ns.
File "src/standard/id.ml", line 1:
Error: Some fatal warnings were triggered (2 occurrences)
      ocamlc .dolmen.objs/dolmen__Id.{cmo,cmt} (exit 2)
```
Fixind that it will compile well.

Gbury:
Is it a bug in dune that the menhir stanza does not work with subdirectories ? or is it expected ?

About Tokens, there should not be any problem, the ml and mli files generated only contain a single type declaration, so unused warning should never trigger on these files. Anyway, the separation of the token declaration in a separate file is there for the tokens to be accessible in the lexer (since the parsers are functorized, the token declaration need to be outside the functor).

Gbury:
With the last commit (using your manually unrolled dune file). I still get the following errors:

```
File "src/dune", line 31, characters 0-176:
31 | (rule
32 |   (targets lexDimacs.ml)
33 |   (deps    src/languages/dimacs/lexDimacs.mll)
34 |   (action  (chdir %{workspace_root}
35 |             (run %{bin:ocamllex} -q -o %{targets} %{deps})))
36 | )
Error: No rule found for src/src/languages/dimacs/lexDimacs.mll
File "src/dune", line 66, characters 0-230:
66 | (rule
67 |   (targets parseDimacs.ml)
68 |   (deps    src/languages/dimacs/parseDimacs.mly  src/languages/dimacs/tokens_dimacs.mly)
69 |   (action  (chdir %{workspace_root}
70 |             (run %{bin:menhir} --explain %{deps} --base parseDimacs)))
71 | )
Error: No rule found for src/src/languages/dimacs/parseDimacs.mly
```

anmaped:
> I guess we'll need to also unroll stanzas to make menhir compile the Tokens files?

Yes. BTW, I'm finishing a modular dune. I will do a commit as soon as possible.

Gbury:
Great !
While you're at it, if you could update the Makefiles rules, and/or move the makefile rule to the root of the project, since it appears dune prefers to be run from there, that would be perfect !

anmaped:
@Gbury Could you give me permissions to write to the dune branch?

anmaped:
I will commit the other modules later since we have several warnings due to the records (some you already fixed in b076742d8b1c68c0a7174d9ab652d996dc69c868). I also have used `--profile release` to surpass the warnings.

Btw, we have an error in ptpt.
```
File "parseTptp.mly", line 572, characters 4-16:
Error: This expression has type ?loc:L.t -> T.t list -> T.t -> T.t
       but an expression was expected of type
         loc:L.t -> 'tv_variable list -> T.t -> T.t
```

Gbury:
Ok, I think we're almost there ! Could you also update the makefile to also build the `dolmen` executable with dune ?
Dune complains, or rather generates,dune-project files in directories interface, languages, and standard, could you commit them ?
Lastly, it seems that most warnings that prevent dune from compiling things without the release profile is actually code generated by menhir, over which we have little control, maybe that would merit a bug report in dune (so that dune disables this warnings on files generated by menhir ?).

anmaped:
> Btw, we have an error in Tptp. I will check if it is some flag.
> 
> ```
> File "parseTptp.mly", line 572, characters 4-16:
> Error: This expression has type ?loc:L.t -> T.t list -> T.t -> T.t
>        but an expression was expected of type
>          loc:L.t -> 'tv_variable list -> T.t -> T.t
> ```

This was due to the infer flag in menhir (I just updated menhir to the last version and now it is working). dune-project file was changed to enforce menhir 2.0.

anmaped:
@Gbury Should we do a transitory wrap of packages or definitive?

Gbury:
Sorry for the delay. Could you explain a bit more what you mean by transitory wrap and definitive wrap ?

Gbury:
Basically, I think the best would be to let dune install all the required files (i.e. having an empty `install` field in the opam file, given that dune generates the required .install and META files): it knows what files to copy, and what conventions to respect. Moreover, it can very easily install sub-libraries like what we'll need for the typechecking and printer sub-packages.

In terms of packaging and libraries I'd say we should aim at the following:
- 3 opam packages : dolmen, dolmen-typecheck, and dolmen-export (or dolmen-pretty ?)
- at least the 3 related ocamlfind libraries (preferably using sub-libraries): dolmen, dolmen.typecheck, dolmen.export
- remains the question of how dune handles all the sub-directories. Since we're going with the traditional dune idiom where each directory corresponds to a findlib library, this means quite a lot of libs (one for each language, etc...). Each of these libraries can be public I'd say (so that people with very specific need can cherry-pick exactly what they need. However, there should still be a main entry point: a lib called `dolmen` (as is currently the case in released version), which depends on all the sub-libs that dune creates, so that the switch to dune does not break everything if possible.

anmaped:
> 
> 
> Basically, I think the best would be to let dune install all the required files (i.e. having an empty `install` field in the opam file, given that dune generates the required .install and META files): it knows what files to copy, and what conventions to respect. Moreover, it can very easily install sub-libraries like what we'll need for the typechecking and printer sub-packages.

Yes, it's true, but dune will create a lot of folders instead of sub-folders in opam lib directory. It's why I have done it manually.
 
> 
> In terms of packaging and libraries I'd say we should aim at the following:
> 
>     * 3 opam packages : dolmen, dolmen-typecheck, and dolmen-export (or dolmen-pretty ?)
> 
>     * at least the 3 related ocamlfind libraries (preferably using sub-libraries): dolmen, dolmen.typecheck, dolmen.export
> 
>     * remains the question of how dune handles all the sub-directories. Since we're going with the traditional dune idiom where each directory corresponds to a findlib library, this means quite a lot of libs (one for each language, etc...). Each of these libraries can be public I'd say (so that people with very specific need can cherry-pick exactly what they need. However, there should still be a main entry point: a lib called `dolmen` (as is currently the case in released version), which depends on all the sub-libs that dune creates, so that the switch to dune does not break everything if possible.

Now we have.
dolmen; dolmen.smtlib, dolmen.standard, dolmen.interface etc...

anmaped:
> 
> 
> Sorry for the delay. Could you explain a bit more what you mean by transitory wrap and definitive wrap ?

Each public library can be wrapped with its name, similar to what ocamlbuild do with (for-pack). However, dune support transitive wrapping and shows some messages to inform the developers that the usage of the library is deprecated.

Gbury:
> Yes, it's true, but dune will create a lot of folders instead of sub-folders in opam lib directory. It's why I have done it manually.

It does indeed work currently, but I suspect that as soon as we try to have multiple opam sub-packages, the Makefiles rules to decide what to install and where will become quite messy. So i would much prefer if installation was left to opam, using the dune generated files.

> Now we have.
> dolmen; dolmen.smtlib, dolmen.standard, dolmen.interface etc...

Currently, pinning the repo to the dune branch installs 4 ocamlfind packages: `dolmen`, `dolmen.line`, `dolmen.smtlib`, and `dolmen.standard`. I suppose installing the remaining libs is a question of adding the right dune files ?

Also, and more worrisome, is that the `dolmen` main library is missing the module aliases one would expect: for instance in order to access the `Logic` class, the `Dolmen.Logic`  module path is not accessible anymore and instead the path is `Dolmenclasses.Logic`. I'd much prefer to have something like `Dolmen_Classes.Logic`, or `Dolmen_classes.Logic`, and have an alias in the `dolmen` package so that `Dolmen.Logic` works (if I understand correctly, the `Dolmen_classes` module is necessary given the split into multiple ocamlfind libraries introduced by dune ?).

> Each public library can be wrapped with its name, similar to what ocamlbuild do with (for-pack). However, dune support transitive wrapping and shows some messages to inform the developers that the usage of the library is deprecated.

Transitive wrapping seems like a good idea, at least for the sub-libraries standard, interfaces and classes (it is reasonable for each language to have its own library, but these could also be transitively wrapped).

anmaped:
> 
> 
> > Yes, it's true, but dune will create a lot of folders instead of sub-folders in opam lib directory. It's why I have done it manually.
> 
> It does indeed work currently, but I suspect that as soon as we try to have multiple opam sub-packages, the Makefiles rules to decide what to install and where will become quite messy. So i would much prefer if installation was left to opam, using the dune generated files.
> 
> > Now we have.
> > dolmen; dolmen.smtlib, dolmen.standard, dolmen.interface etc...
> 
> Currently, pinning the repo to the dune branch installs 4 ocamlfind packages: `dolmen`, `dolmen.line`, `dolmen.smtlib`, and `dolmen.standard`. I suppose installing the remaining libs is a question of adding the right dune files ?

It's a matter of modifying the META file.
> 
> Also, and more worrisome, is that the `dolmen` main library is missing the module aliases one would expect: for instance in order to access the `Logic` class, the `Dolmen.Logic` module path is not accessible anymore and instead the path is `Dolmenclasses.Logic`. I'd much prefer to have something like `Dolmen_Classes.Logic`, or `Dolmen_classes.Logic`, and have an alias in the `dolmen` package so that `Dolmen.Logic` works (if I understand correctly, the `Dolmen_classes` module is necessary given the split into multiple ocamlfind libraries introduced by dune ?).

Yes. I guess that doing Dolmen.Logic is impossible now with dune. We have two options, Dolmen_classes or Logic.  Indeed, I tried to put this library private, but the dependencies break. I have to check it again.


> 
> > Each public library can be wrapped with its name, similar to what ocamlbuild do with (for-pack). However, dune support transitive wrapping and shows some messages to inform the developers that the usage of the library is deprecated.
> 
> Transitive wrapping seems like a good idea, at least for the sub-libraries standard, interfaces and classes (it is reasonable for each language to have its own library, but these could also be transitively wrapped).

Right.

Gbury:
> > > Yes, it's true, but dune will create a lot of folders instead of sub-folders in opam lib directory. It's why I have done it manually.
> > 
> > 
> > It does indeed work currently, but I suspect that as soon as we try to have multiple opam sub-packages, the Makefiles rules to decide what to install and where will become quite messy. So i would much prefer if installation was left to opam, using the dune generated files.
> > > Now we have.
> > > dolmen; dolmen.smtlib, dolmen.standard, dolmen.interface etc...
> > 
> > 
> > Currently, pinning the repo to the dune branch installs 4 ocamlfind packages: `dolmen`, `dolmen.line`, `dolmen.smtlib`, and `dolmen.standard`. I suppose installing the remaining libs is a question of adding the right dune files ?
> 
> It's a matter of modifying the META file.

How hard would it be the let dune generate the META file ? Mainly because, again, the main goal of switching to dune was to simplify the installation of opam (and findlib) sub-packages for printers and typecheckers, so if the `Makefile` still does the job of deciding what to install and where, the switch isn't that useful.

> 
> > Also, and more worrisome, is that the `dolmen` main library is missing the module aliases one would expect: for instance in order to access the `Logic` class, the `Dolmen.Logic` module path is not accessible anymore and instead the path is `Dolmenclasses.Logic`. I'd much prefer to have something like `Dolmen_Classes.Logic`, or `Dolmen_classes.Logic`, and have an alias in the `dolmen` package so that `Dolmen.Logic` works (if I understand correctly, the `Dolmen_classes` module is necessary given the split into multiple ocamlfind libraries introduced by dune ?).
> 
> Yes. I guess that doing Dolmen.Logic is impossible now with dune. We have two options, Dolmen_classes or Logic. Indeed, I tried to put this library private, but the dependencies break. I have to check it again.

I don't see why it isn't possible. As it stand, the `dolmen` findlib package depends on the others (`dolmen.line`, etc...), right ? in That case, the top-level `dolmen` lib can simply alias these modules ? (dune allows to have a dolmen.ml top-level file containing the manual aliases does it not ?).

anmaped:
> I don't see why it isn't possible. As it stand, the dolmen findlib package depends on the others (dolmen.line, etc...), right ? in That case, the top-level dolmen lib can simply alias these modules ? (dune allows to have a dolmen.ml top-level file containing the manual aliases does it not ?).

No. dolmen is the small unit that you can use, the other packages are the extensions that depends on dolmen. Of course, if you want to do a manual alias, it is possible, but again, you are decentralizing the build and now centralizing the usage.

anmaped:
We only have to see if we could generate the META file automatically.

Gbury:
So I fixed the META file generation, as well as letting dune handle all of the installation, and the wrapping of the core sub-libraries into the main `dolmen` lib.

anmaped:
> 
> 
> So I fixed the META file generation, as well as letting dune handle all of the installation and the wrapping of the core sub-libraries into the main `dolmen` lib.

The dots were the trick for META files. I didn't know that.

However, dune is not installing the files right now and there is something missing in opam.

anmaped:
I confirm it's working.

PR #18, opened 2020-02-14
Add versioned parser for languages

Gbury:
This PR is (currently) a prototype to look at versioning the parsers in dolmen, so as to be able to have a parser for multiple versions of a language.

Currently, this was done by versioning the current tptp parser (which corresponds to v6.3.0).

The goal is to have sub-libraries (e.g. `dolmen.lang.version`) for each supported version of a language, and the main lib for a language (e.g. `dolmen.lang`) would point to the latest version.

cc @c-cube


c-cube:
I think it's nice to version parsers. Funnily, I'd have started with smtlib (2.6) :-)

Gbury:
All the parsers will be versioned, this is more of a POC to discuss the actual implementation of how the parsers are versioned, ^^

src/languages/tptp/dolmen_tptp.ml
```diff
@@ -1,19 +1,9 @@
...
-    (T : Term with type location := L.t and type id := I.t)
-    (S : Statement with type location := L.t and type id := I.t and type term := T.t) =
-  Dolmen_std.Transformer.Make(L)(struct
-    type token = Tokens_tptp.token
-    type statement = S.t
-    let env = ["TPTP"]
-    let incremental = true
-  end)(LexTptp)(ParseTptp.Make(L)(I)(T)(S))
+(* Include the latest version *)
+include Dolmen_tptp_v6_3_0
```
c-cube: rather than `include`, could it be `module Current = V6_3_0`? keeps things a bit tidier.

Gbury: It could, but that would break compatibility for the current users...

c-cube: ah, ok. but then I think the entrypoint for `V6_3_0` should be elsewhere so if you use the versioned API you don't see all the clutter from the current version.

Gbury:
So this one should be ready to review if you're motivated @c-cube , else I'll merge this week-end.

Issue #26, opened 2020-02-21
smtlib parsing: assertion hit when parsing wrongly typed equation


quicquid:
Please consider the following smtlib script:

```
(set-logic AUFLIA)
(declare-fun notindomain (Int) Int)
(assert (forall ((i Int))
                (= (notindomain i) (or (< i 0) (< 5 i)) )))
(check-sat)
```

It contains a type error because notindomain should be of type Int to Bool. When I run dolmen on the file, I receive the following error message:

```
Error While typing: (= (notindomain (i)) (or (< (i) 0) (< 5 (i))))
File '/tmp/test.smt2', line 6, character 16-57:
The term: (Fatal error: exception File "src/standard/expr.ml", line 484, characters 6-12: Assertion failed
```

The error occurs in pretty printing (the assertion requires an infix operator to have at least 2 arguments) so it might hide the actual error.

Gbury:
This was solved by PR#27 but apparently github didn't automatically close the issue, ^^

PR #27, opened 2020-02-28
Fix issue 26

Gbury:
This should fix issue #26 .

Note however, that the error message is not as one could expect: this is because currently equality between propositions is allowed and typechecked into an equivalence. I will check whether this matches the standard of tptp and smtlib; is it allowed in zf @c-cube  ?


c-cube:
I need to check the grammar, but if there's no token for `<=>` then it should definitely work :p

PR #33, opened 2020-04-18
[Pipes] return all the possible statements

### LINKED ISSUE #32: Interface of Pipes
The interface of `Pipes` is at the same time in control through the function of the State module and give-up control through `Continue`. Is it really needed to have this mixed instead of one or the other or both but separated? I can propose an MR to make things more clear.

bobot:
Fixes #32


Gbury:
Great, thanks a lot !

PR #34, opened 2020-04-18
[smtlib_logic] Parse part by part instead of globally

bobot:


Gbury:
Small note: I might later need to add some special casing for some logics with particularly annoying (and historic iirc) restrictions, such as `QF_ABV` only allowing some arrays from bitvectors to bitvectors, but the generic approach in this PR will make these special cases much more readable and manageable, ^^

Gbury:
I'd be very happy if Dolmen became the reference for the SMTLIB, :)
About SMTLIb v3, I don't know what timeframe (i.e. when is it planned to be released), but I should have the time to add support for it reasonably quickly I think. In fact, a lot of my work recently on dolmen was to properly version the smtlib support, with the explicit goal of making the addition of SMTLIb v3 easier. I'll try and write an email to Cesare, Pascal and Clark about that (as well as the questions we've come across).

Gbury:
Now, concerning the functions symbols, and so on, the distinctions I see is the following: constants are simply functions of arity 0, so I don't make much of a distinction between constants and functions, however, what seems to matter more if the `uninterpreted` part of declaring (or defining) new constants/functions, which in essence boils down to the ability to declare new sorts (which are always uninterpreted when they are not part of a theory). Indeed, forbidding to declare any constants seems like a very heavy restriction

So I think we can have the following distinctions:
- declaring of new (thus uninterpreted) sorts can be allowed/disallowed depending on some restrictions/flags
- declaring of constants/functions at the term level can be split in two:
  - interpreted: when the type of the declare constant only contains sorts coming from theories, it can be considered to interpreted. In this case, an arrow/function type is actually considered uninterpreted (so there could actually be a difference between functions and constants).
  - uninterpreted: if the type of the constant being declared contains any uninterpreted sort (including function arrows and/or declared uninterpreted sorts), it would be uninterpreted.

Currently, dolmen always allows to declare new (uninterpreted) sorts and constants/functions, but preventing some of these (according to the distinctions above), is pretty easy.

With all that, the `UF` theory would allow declaring of uninterpreted new sorts and constants/functions (in this case, it would become a theory rather than a restriction), but we still have the following questions:
- it's not clear from the descriptions which logics in SMTLIB include the  `UF` theory, or rather which do not
- are there theories that disallow declaring new interpreted constants ?

Gbury:
It's good for now, I'll merge and add handling for Uf shortly.

PR #35, opened 2020-04-21
Term_def gives the expr id given to the defined symbol

bobot:
Types definition are substituted but Term definition are not so we need the `Expr.term_const` associated.

### DIFF
--- a/src/loop/pipes.mli
+++ b/src/loop/pipes.mli
@@ -42,9 +42,9 @@ module Make
 
   type defs = [
     | `Type_def of Dolmen.Id.t * Expr.ty_var list * Expr.ty
-    | `Term_def of Dolmen.Id.t * Expr.ty_var list * Expr.term_var list * Expr.term
+    | `Term_def of Dolmen.Id.t * Expr.term_const * Expr.ty_var list * Expr.term_var list * Expr.term
   ]
-  (** The type of top-level type definitions. *)
+  (** The type of top-level type definitions. Type definitions are inlined and so can be ignored. *)
 
   type assume = [
     | `Hyp of Expr.formula
--- a/src/loop/pipes.ml
+++ b/src/loop/pipes.ml
@@ -32,7 +32,7 @@ module Make
 
   type defs = [
     | `Type_def of Dolmen.Id.t * Expr.ty_var list * Expr.ty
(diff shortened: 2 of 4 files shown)


src/loop/pipes.mli
```diff
@@ -42,7 +42,7 @@ module Make
 
   type defs = [
     | `Type_def of Dolmen.Id.t * Expr.ty_var list * Expr.ty
-    | `Term_def of Dolmen.Id.t * Expr.ty_var list * Expr.term_var list * Expr.term
+    | `Term_def of Expr.term_const * Expr.ty_var list * Expr.term_var list * Expr.term
   ]
   (** The type of top-level type definitions. *)
```
bobot, line 47: @gbury Is it true?
```suggestion
  (** The type of top-level type definitions. Type definitions are inlined and so can be ignored. *)
```

Gbury: Yes it is correct.

src/loop/pipes.mli
```diff
@@ -42,7 +42,7 @@ module Make
 
   type defs = [
     | `Type_def of Dolmen.Id.t * Expr.ty_var list * Expr.ty
-    | `Term_def of Dolmen.Id.t * Expr.ty_var list * Expr.term_var list * Expr.term
+    | `Term_def of Expr.term_const * Expr.ty_var list * Expr.term_var list * Expr.term
```
Gbury, line 45: I'd prefer if this was instead ``` `Term_def of Dolmen.Id.t * Expr.term_const```, this way the information about the dolmen identifier is not lost (keeping it will prove useful for the upcoming export function).

bobot: I have nothing against, but it is not kept for ` `Term_decl`, so what is the rational? Should it not be put inside the `Expr.id`, an `origin` field? Finally what is export function? :smiley_cat:

Gbury: That's a good point for ``` `Term_decl ``` , I'll have to think about it a bit more (in any case, as long as there's no release the change isn't really a problem).  So you can leave your change as-is.

The export function, or rather functionality is simply to have dolmen be able to output the typed problem in a chosen format (if possible, since not all formats support the same theories). I'm planning to add that soon-ish (I already have a big part of the code in archsat, which is capable of doing this kind of conversion).

PR #36, opened 2020-04-27
Implement UF/DT theory restrictions

Gbury:
This is still WIP, but contains quite a few changes:
- the typechecker and typing loop are adapted to reject declarations that do not respect the restrictions imposed by the (lack of) UF/DT theories
- some additions to error reporting which is missing a few cases (still in progress)
- the representation of smtlib logic is a bit more structured now, restrictions are transformed into a set of features represented as field of a record
- the parsing of smtlib logic is made a bit more strict by taking into account:
  - the order in which theories can appear in a logic name
  - the fact the a theory can not be repeated in a logic name

This last point probably some review/comment by @bobot , who should be more knowledgeable about whether it is correct/reasonable; the idea was to not allow to parse logic such as `LIABVQF_ANIRA`, which the current code happily parses, even though it's not really a valid theory declaration (it's incoherent with respect to arithmetic for instance).

### DIFF
--- a/dune
+++ b/dune
@@ -1,0 +1,3 @@
+; Ensure no inlinging takes place in dev mode to have more accurate backtraces
+(env (dev (ocamlopt_flags -g -inline 0)))
+
--- a/src/loop/state_intf.ml
+++ b/src/loop/state_intf.ml
@@ -13,7 +13,7 @@ type source = [
   | `Raw of string * string
 ]
 
-type phase =[
+type phase = [
   | `Parsing
   | `Include
   | `Typing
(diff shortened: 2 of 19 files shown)


Gbury:
Well, I mostly looked at all the logics listed on http://smtlib.cs.uiowa.edu/logics.shtml , which contains `QF_ABV` , `QF_AUFBV`, `QF_ALIA` and `QF_AUFLIA`, so it seems that the array theory can indeed occur after a `QF_`, or did I miss something ?

Gbury:
Ok, so this should be fine to merge ?

Gbury:
I'll merge, but don't hesitate to open issues if there are more logics that should be supported.

PR #52, opened 2020-09-23
error message for syntax

c-cube:
can't figure out how to test it, now I get:

```sh
$ dolmen foo.smt2
File "", line 1, character 104-105:
Error Syntax error
```

which is weird, for input:

```
(set-logic ALL)
(declare-datatypes ((t 0)) (((A) (B) (C) (D))))
(define-fun f ((x t)) t (match x ((A B) (_ x))))
(check-sat)
```

### DIFF
--- a/src/bin/errors.ml
+++ b/src/bin/errors.ml
@@ -69,8 +69,8 @@ let exn st = function
         (if l.Dolmen.Std.Loc.start_line = 1 &&
             l.Dolmen.Std.Loc.stop_line = 1 then prelude_space st else "")
         Dolmen.Std.Loc.fmt_hint l;
-    Loop.State.error ~loc: { file; loc; } st "%s@."
-      (match msg with "" -> "Syntax error" | x -> x)
+    Loop.State.error ~loc: { file; loc; } st "%a@."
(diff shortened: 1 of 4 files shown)


Gbury:
Yes, the location is a bug I'm currently trying to fix. The error message not being used is a problem though, I'll take a look at it as soon as possible.

Gbury:
It's possible (though i'm not sure yet), that the message you added is not for the correct situation (I think the error case where you added the message has one parenthesis more than the example); I'll check in more details tomorrow.

c-cube:
this one seems to work. might have overshot a bit :grin:

Gbury:
Could you add a regression test for that (either in `tests/regression/pr52`, or in `tests/parsing/smtlib/v2.6/`) ? :D

Gbury:
Ok, so I fixed the location error in 5be9ce56d8fa206deb99491d1dec818274fa5d93 (and also added a test to check syntax errors, which happen to be exactly your test case, ^^). You can rebase, run `make test`, and then `make test-promote` to update the expect file for the test, and then I'll merge the PR, :)

Gbury:
Nevermind, i did it, ^^

Gbury:
Thanks a lot !

Issue #48, opened 2020-07-06
Smtlib v3


bobot:
The proposal is becoming more tangible http://smtlib.cs.uiowa.edu/version3.shtml. I believe it could give good feedback to try to implement it in dolmen. I can give a hand, but I would certainly need your help @Gbury to know the best way to encode namespace and filters for example. Do you think it is a difficult task?

Gbury:
Typing smtlib3 is going to be a non-trivial task indeed. I've already started to think about the changes that would be required to properly handle it. Currently the main points that will requires changes are:
- modules (this is not really that complex, rather it requires a bit of thinking about the design and import/open/lookup complexity tradeoffs in terms of performance)
- overloading of user-defined operators: this is going to be annoying as it will most certainly require to support overloading in the core typechecker (rather than support it in the builtins as done currently), and I think it might require the typechecker to inspect (or rather unify) types in order to be able to resolve a name to the correct typed symbol, which I don't like because it basically breaks the abstraction between the typer and the expressions, :/
- qualified imports with restrictions to only some instances of polymorphic type constructors: as far as I can tell this is the real problem, and might require a lot to correctly work, depending on the exact specification of such import restrictions (e.g. what happens if functions imported by this module can produce some values of types not imported, would this have to be checked and rejected by the type-checker or would it be the user's responsibility to import a coherent subset of a module, what happens with imports from other modules which can produce values of not imported types, what happens if the same module is imported multiple times with different type instances allowed, etc...)
- upgrading the Expr module (and the core typechecker) to higher-order (this one should be fairly easy).

So quite a lot of non-trivial work. I'll try and work on that as soon as possible. Do you have any idea of the timeframe of the v3 release ?

bobot:
At least 1 year for a final version, I believe. I think you could express your difficulty on the SMTLIB mailing list if you think that some part are too complicated or not precise enough. For example overloading can be accepted only if an `as` is provided (and it is now with the correct type of the function not only the result) or all the argument have a type without non-user type variable. Or overloading of polymorphic function can be completely forbidden: perhaps polymorphism or overloading not both, for defining the family of a particular function symbol!

At the end it is just a first public draft. Once available you should listen to @tinelli talk at SMTLIB: one important point is that it should not be a too big burden for solver implementor. So you can choose your own restrictions if you think they make sense and they help you to implement a possible/simpler v3, then we can propose them as amendment to the draft.

Gbury:
[...]

PR #53, opened 2020-10-16
[RFC] Add polymorphism of smt v3 in smt v2.6

bobot:
Over the year adding polymorphism to the smt language have been proposed a lot of time. It will become standard in v3 as already discussed in #48. But for COLIBRI I needed to be able to parse polymorphism (since it knows little above ground term any encoding is problematic). Finally the addition is quite small since dolmen already handles it internally.

Currently the modificatio is as simple as possible:
  - In v2.6
  - Without a flag to restrict it

In order to solve that, and if you are interested, what is the right way since v3 should be mainly backward compatible with 2.6:
 - A copy of v2.6 as v3?
 - Just add a new langage but still share the code?

I don't plan to add a v3 parser now, first because v3 is still not formalised and because I believe it is be better to do it step by step.


Gbury:
If I understand correctly, this is a compatible extension of smtlib v2.6, correct ?
In any case, i have a few remarks:
- considering dolmen may be used to check new benchmarks for the smtlib, this feature would either need to be protected by a flag (or be only in alternative version for smtlib)
- thinking about it, I think having the extension be a new smtlib version (we'd need to find a name though, :p), would be the nicest from the point of view of users of dolmen. It could be the default for file with the `.smt2` extension, but if needed an explicit set of the input language to smtlibv2.6 would be all that is needed to ensure user that what is parsed is exactly smtlib2.6
- I know that @OCamlPro-Coquera also worked on an extension of smtlib v2.6 to support polymorphism, called psmt2, it'd be nice to at least compare the two approaches, and ideally have them be the same thing (since psmt2 is planned to be included in dolmen at some point)

Finally, note that I recently added error messages to smtlib, though it seems I didn't correctly set up testing to fail on syntax change, I'll fix that in master (and then you'll have the fun of writing the error messages for the new cases (and potentially fixing some of the already existing ones)).

Gbury:
Actually, the test works fine (just run `make test`), but the question of error messages can come after we decide where the extension will live.

Gbury:
@bobot would the psmt2 format work for you (until smtlib3 is released) ?

Gbury:
If everyone can be happy with it, I think the best thing to do would be to add a new version of smtlib2 that contains the required extensions for polymorphism, probably called `psmt2` if everyone agrees, since the name already kind of exists. If I remember correctly, psmt2 has a dedicated different extension `*.psmt2`, which would help differentiate from regular smt2 files. Would that work for you @bobot ?

Gbury:
I'll start on that, and ping you once you can try it out.

src/languages/smtlib2/v2.6/parseSmtlib.mly
```diff
@@ -248,11 +248,16 @@ datatype_dec:
 
 function_dec:
   | OPEN s=SYMBOL OPEN args=sorted_var* CLOSE ret=sort CLOSE
-    { I.(mk term s), args, ret }
+    { I.(mk term s), [], args, ret }
+  | OPEN s=SYMBOL PAR OPEN vars=datatype_symbol+ CLOSE OPEN args=sorted_var* CLOSE ret=sort CLOSE CLOSE
+    { I.(mk term s), vars, args, ret }
```
Gbury, line 253: In this new production rule (and for the function definition), you put two CLOSE at the end, which makes the rule not well-parenthesized. From what I can tell, it's probably a mistake that can be solved one of two ways:
- have only one CLOSE parenthesis at the end
- require an OPEN before the PAR

Gbury:
So, I just pushed on the repo a branch `psmt2` which adds a `poly` version for the smtlib2 language (recognized by the `.psmt2` extension), which is exactly your polymorphic extension of the smtlib2 language (which luckily coincides with the `psmt2` format as created by the alt-ergo developpers, yay !).

From now on, I'd suggest you do the following:
- set your branch for this PR to the head of the `psmt2` branch to continue work until things wan be merged
- complete the syntax error messages of the `smtlib2/poly` language (see https://github.com/Gbury/dolmen/blob/master/HACKING.md#hand-written-syntax-error-messages for info about how to do that; and if the explanations there are not clear/informative enough, don't hesitate to tell me and I'll try and complete things): this means adding cases for all errors in the `syntax.messages` file, and adding corresponding test cases as described in the `HACKING.md` file
- add some good test files (as in tests that pass and do not raise an error), for both parsing and typing (resp. in ` tests/parsing/smtlib/poly/` and `tests/typing/base-smtlib/poly`)

And then we should be good to go, :)

src/languages/smtlib2/v2.6/parseSmtlib.mly
Gbury: I choose the route of adding an `OPEN` token before the `par` as that seemed more coherent with how the rest of the extension works; see branch `psmt2`.

Gbury:
Great ! (Though just to be clear,rather than setting the target of your PR to the `psmt2` branch, what I was suggesting was that you set your branch to point at the current HEAD commit of `psmt2`, still targeting `master` with the PR, and continue working from there (since the branch contains all your changes applied to the new parser for psmt2 rather than smtlib2))

Gbury:
> What is the version of `lsp` that should be used?

The latest release (`1.2` iirc) should work (previous releases have a bug related to vendored libs which makes it impossible to use in conjunction with uutf).

Gbury:
> * The tests framework of dolmen can be added as a datapoint to [ocaml/dune#3901](https://github.com/ocaml/dune/issues/3901) , and it has the advantage to be public.

I'v just posted on the issue to describe how and why dolmen generates dune files.

Gbury:
> I have a compilation error with `1.2.0` of lsp, and the travis job is failing for another reason.

I'll take a look, ^^

src/loop/parser.ml
```diff
@@ -48,7 +48,7 @@ module Pipe
       | `Stdin ->
         let lang, file_loc, gen, _ = Logic.parse_input
             ?language:(State.input_lang st)
-            (`Stdin (Logic.Smtlib2 `Latest))
+            (`Stdin (Option.value ~default:(Logic.Smtlib2 `Latest) (State.input_lang st)))
```
Gbury, line 51: This should not be necessary in theory.
The `Logic.parse_input` function already takes the `?language` argument value over the ``` `Stdin l ```if it is present.

Gbury:
> * Updating `syntax.messages` is not trivial from `new.messages`: perhaps I used the wrong method (`meld`) or it could be improved in menhir (updating and adding new states).

I know, however, it seems to be by design of menhir (see https://gitlab.inria.fr/fpottier/menhir/-/issues/14 for more information)

src/loop/parser.ml
bobot: Without that I'm not able to do `cat foo.psmt2 | dolmen -i psmt2` . The value of language is not used for the returned language:

```
  let parse_input ?language = function
    | `Stdin l ->
      let _, _, (module P : S) = of_language
          (match language with | Some l' -> l' | None -> l) in
      let locfile, gen, cl = P.parse_input `Stdin in
      l, locfile, gen, cl
```

bobot: (For another MR, but can we have that `dolmen -i psmt2 -` reads stdin but not in interactive mode? or another way to deactivate interactive mode for stdin?)

Gbury: Oh, I see, the problem is actually that this portion of code should bind the returned language from the `of_language` call, and return that instead of `l`, this should fix things.

For the read stdin but not in interactive mode: sure (and indeed in another PR/MR). I'll just have to think about how that interacts with the already present `mode` option (whether there needs to be another option or if a single option is enough/better).

bobot: Ok fixed in the right way

Gbury:
> > I have a compilation error with `1.2.0` of lsp, and the travis job is failing for another reason.
> 
> I'll take a look, ^^

Okay, so lsp actually broke the API compatibility between `1.1.0` which had the vendor problem, and `1.2.0` which as far as I can tell, doesn't install all the necessary libraries (see https://github.com/ocaml/ocaml-lsp/issues/311 ). It'll likely take some time to fix, so you can ignore the lsp part for now, ^^

Gbury:
For the lsp status: I've added in `master` a pin-depends to a working version of the `lsp` package, with that version you should be able to build everything.

Gbury:
Okay....

So, first let me sum up the few rules I'd like to enforce in dolmen, and then let's talk about what we can do, ^^

I very much want to keep the fact that for official versions (i.e. smtlib v2.7, the incoming smtlib v3, etc..), dolmen parses and types exactly what is documented/specified by the language. Additionally, since parsers are not trivial things, I'd prefer to keep a "clean" parser for these official versions.

For smtlib3 I'd prefer to wait for a (even alpha/beta) version of the language before adding it to dolmen (in any case, there are a few big changes to make in dolmen before smtlib3 can be added, e.g. higher-order and modules, which i'm working on).

For other versions of the language, in particular extensions, i'm much more relaxed on what we can accept[1]. So there can be a lot of solutions:
- if the various polymorphic extensions can be parsed with the same parser (i.e. the extended parser can incorporate all parsing rules without any conflict), and the extensions authors agree, we could keep parse them all with the same parser and label that extension as `.psmt2`
- if the author's extensions agree, maybe one of the extensions could be deprecated in favour of another ? (cc @ACoquereau @rbonichon ), and then dolmen would support the non-deprecated extension
- if not, then we could have a version for each extension (so `.psmt2` following alt-ergo's extension, and another one for 
your extension), and if you have a way to share parsers/error message between those versions, that could be interesting.
- another solution ?

So basically, i think the one question to answer first would be: can we have a single menhir parser that accepts all (or most) of the extensions without any conflict ?

[1] the only restriction that I'll enforce is that automatic extension detection should go for official versions, i.e. extensions of smtlib2 can either:
- use the `.smt2` extension but then the exact language version will have to be given on the command line for the dolmen binary
- or use another extension, such as `.psmt2`
In short the `.smt2` extension will by default always mean the latest official smtlib2 version.

ACoquereau:
> all different for `define-fun`:
> 	start 	stop
> 1. 	before symbol name 	after body
> 2. 	after symbol name 	before body
> 3. 	after symbol name 	after body
> 
> The second should disallow the use of the type variable in type ascription `(AS ...)` in the body, which seems very restrictive. I would prefer to add to dolmen the last one, but the extension `psmt2` seems wrong since it is the name of the second one. `.p2smt2`? `.smt3`?

The way psmt2-frontend handle `define-fun` is wrong, like you said @bobot we should not be able to use type variable in the body. But we did this to have a generic form of `define-fun` with `define-fun-rec` and `define-funs-rec`. I agree that the best way to handle define-fun is the third one, but what are the solutions for `define-funs-rec` ?
```
(define-funs-rec
  (f (par (A) ((x A)) Bool))
  (g (par (B) ((y B)) Bool))
     ((g x)
      (f y)))
```

ACoquereau:
>     * if the author's extensions agree, maybe one of the extensions could be deprecated in favour of another ? (cc @ACoquereau @rbonichon ), and then dolmen would support the non-deprecated extension

I've tested to change the parser of psmt2-frontend and it can handle both the solution 2 and 3 at the same time. I think that the best thing to do for me is to depreciate the solution 2 (but keep it available for benchs compatibility) for psmt2-frontend and add support for solution 3. 

By depreciating the solution 2 the set of benchs generated in 2018 are not relevant anymore. This benchs were generated with Why3 and if i understand correctly, @bobot you made a MR with a printer in why3 able to produce the third solution ? If so we could regenerate our benchs with the new format with de extension psmt2.

rbonichon:
I'm not sure the polymorphic notation we introduced back then (for veriT) needs to be retained. Pretty sure that smt3 notation will be adopted.

Gbury:
> The trigger with `as` are not well taken into account (last problem from the psmt2 benchmark), but I will let you see the problem, I dont' understand how attributes must be handled.

Can you send me the problematic file ?
I agree that attributes are not the simplest part of dolmen, I'll get on writing some doc about that one of these days, ^^

Gbury:
Ok, so the problem you encounter with "as" is the following: there is a `:pattern` annotation, which is meant to add a trigger annotation to a term and has as payload an s-expression (which is not parsed the same as terms according to the smtlib doc). This s-expression is understood as a list of terms, even though syntaxically, we have (recursively), a list of s-expressions and **not** a list of terms. Currently the typechecker correctly handles that, except that since s-expressions do not have the same syntax rules as terms, it does not have the specific parsing rules for `as` and `par`. You did add some production to parse these, but the way the productions work is not correct.

So, basically, according to a strict interpretation of the smtlib standard, `par` and `as` shouldn't really be allowed in these pattern annotations; also note that other constructions such as indexed identifiers (e.g. `(_ bv32 13)` ) are also not allowed technically, which may or may not be annoying in the long term. However, since we are talking about an extension and not smtlib v2.7, we have some leeway thankfully, ^^
In this case, i'd suggest you manually add the required production rules for the `as` and `par` keywords to the rules for parsing s_expressions, copying how they are handled in terms (so I think re-using the already existing  `qual_identifier` production rule instead of the symbol one to handle the `as` keyword, and using the `T.par` function instead of creating a new id with the name `par` to correctly handle `par`).

Gbury:
I'm not quite sure what you'd want, so let me resume how attributes are defined by smtlib2 and handled by dolmen.

In smtlib2, attributes have the form `:name payload?`, where the payload is optional and is an s-expression (whose definition is different from terms). Dolmen parses that as defined by the smtlib2, by adding in the "attrs" field of the untyped term a term of the form `name(payload)` (i.e. the attribute name applied to the payload). Additionally, in s-expressions, application (or rather the recursion production), is not parsed the same as term application, but rather a s-exp of the `(a b c ..)` is parsed as an untyped term of the form `$data(a, b, c, ...)` (i.e. application of a special identifier `$data` to the list of s-expressions). That's for parsing.

For typing, the typechecker does try and type all attributes that are attached to each term (the "try" here means that by default an error in parsing an attribute will only result in a warning and not an error). Since there are close to no attributes defined in the tff typechecker, the typing of an attribute (and its optional payload), will almost always then be routed to the typing builtin function. That allows for each language to define its own typing of attributes. In the case of smtlib2, you can see the attributes defined there :
https://github.com/Gbury/dolmen/blob/dfbd7ff21281d501e145471e7b0d051d8c7116d9/src/typecheck/core.ml#L129-L142

The first two cases define the typing of `:named` and `:pattern` attributes. In the case of patterns, we parse the payload as a list of terms to be added as triggers, using this aux function:
https://github.com/Gbury/dolmen/blob/dfbd7ff21281d501e145471e7b0d051d8c7116d9/src/typecheck/core.ml#L83-L90

And finally, we have the following case which enables the typechecker to parse s-expressions as terms:
https://github.com/Gbury/dolmen/blob/dfbd7ff21281d501e145471e7b0d051d8c7116d9/src/typecheck/core.ml#L144-L151

As such, parsing attributes works for all attributes (that's pretty much the only solution), and typing of these is very easily extensible and adaptable to each language.

However, note that currently, attributes are meant to be non-crucial information attached to terms, or said differently, the presence or absence of an attribute should not change whether a term type-checks; attributes are used as meta-data and/or hints for the consumer, which is why the current use of attributes are for triggers and naming things for later model output. information that is required to succesfuly type or prevent the typing of terms should not be in attributes but as normal nodes of the AST. This is the case for `as` typing annotations, which are translated using the `Colon` juxtaposition node of the untyped AST.

So now that the usage of attributes in dolmen is a bit clearer, we can talk about how to handle uses of `as` in s-expressions, as it happens in the case of typing the payload of a `:pattern` attribute. I find it a bit annoying that smtlib2 (and by extension psmt2) would define payload of attributes as s-expressions to then try and use them as terms. This creates a bit of a problem in my opinion, and will unnecessarily complicate most implementation in my opinion.
So currently, and according to the definition of s-expressions in smtlib2, keywords such as `as` and `par` are **not** allowed in s-expression (i.e. these would be syntax errors, not typing or interpretation errors). It seems that accepting these is however required when users want to use polymorphism in triggers, and thus be able to give terms as payloads. And it's not really clear to me how to handled this in dolmen: as you said, extending the parsing rules for s-expressions to include `as` and `par` feels a bit ad-hoc, on the other hand, since they are keywords, we'll need specific rules to parse them in s-expressions anyway...

Gbury:
Basically, I'd say smtlib2 (and extension's authors) should decide whether attribute payloads are terms or s-expressions (or both, given an adequate syntax rule to distinguish the two cases), and stop trying to say different things depending on the situation.
For instance, the documentation defines in the syntax attributes payloads as s-expressions, but the documentation also says that `:pattern` payloads should be `terms`, and these two statements are contradictory.

Gbury:
> I agree fully with your last comment. And that's why I think it would be simpler if dolmen could parse the `:pattern` directly as a term. However it should be done without making `:pattern` a keyword.

It would be a solution, but that's not really possible with menhir as far as I know, ^^ (plus it would be a hack in the sens that other attributes could not really do the same, except if the mechanism was somehow extensible at runtime, which would be .. complicated ; that's one of the reasons dolmen is likely never going to parse the Coq language).

By the way, I submitted a topic about that in the SMTLIB google group (it's pending for authorization currently).

Gbury:
Sorry for not merging sooner, I'll try and merge asap.
Good catch for the `to_int`/`floor` problem in arith.

However, concerning the second one (about the timeouts), I'm not really sure about its usefulness. I'll need to add it in comments and in the docs, but there is already a way to disable the timer part of the limits, by using a timeout of 0 (see the documentation for Unix.itimer). For the size computation limits, is there a real and measurable cost to the Gc alarm (as in a test case where you can measure/observe the overhead of the alarm when setting the memory limit to infinity) ? And actually, what is the motivation behind the change ? Does it try and solve a concrete problem or is it more for code cleanliness ?

Gbury:
I Have a plan to take care of the patterns properly, i'll try and work on it soon (probably next week).
Actually, I think I'll going to merge this PR as soon as I can, and you'll be able to open other PRs (or open issues) for the minor bugs you mention.

Gbury:
I rebased on master

Gbury:
Merged, thanks a lot for the work !

PR #60, opened 2021-03-24
vendor linol

c-cube:


Gbury:
With linol vendored, should `linol` and `linol-lwt` still be in the deps of the opam package for `dolmen-lsp` ? In any case, I'd expect they should be removed from `pin-depends`, no ?

c-cube:
linol has actually been released! :man_dancing:

PR #62, opened 2021-03-30
Better CI

Gbury:
Will likely fail because of changes to the way opam handles compiler options for switches for ocaml `>= 4.12`


smorimoto:
Inside setup-ocaml, for better reusability, the cache is only created in a form that can be shared across all workflows. Therefore, it's better not to set a cache prefix to make the cache more reusable. If you accidentally cache in the wrong form, you can set it and overwrite the key to avoid the problem.

PR #66, opened 2021-04-14
Add higher-order type-checking to Dolmen

Gbury:
Big changes needed to introduce higher-order, mainly:
- extend the `Std.Expr` module to accommodate higher-order types and terms
- extend the `Std.Expr` module to cleanly deal with type aliases
- extend the type-checker to adequately split arguments of higher-order

This allows to support the whole zf language, and is also a preparation for the upcoming smtlibv3 which will include higher-order.

Even though I tried to minimize them, there are quite a lot of breaking changes, so any feedback is welcome (particularly about the new representation of terms), cc @bobot @c-cube 

Some more tests will come soon, but the PR is already in a good and working state.


src/standard/expr.mli
```diff
@@ -16,736 +16,99 @@ type hash = private int
...
+  mutable alias : type_alias;
+}
 
-(** {3 Types} *)
+and type_alias =
+  | No_alias
+  | Alias of {
+      alias_vars : ty_var list;
+      alias_body : ty;
+    }
```
bobot, line 46: Why `type_alias` was needed?

Gbury: It's not strictly related to higher-order, but since I'm planning to work on printing back problems right after, keeping type aliases rather than expanding them during type-checking will allow to be able to read some smtlib problem containing type aliases, and then print it back, including the type aliases.

src/standard/expr.mli
```diff
@@ -16,736 +16,99 @@ type hash = private int
...
-type let_kind =
-  | Parallel
-  | Sequential
+(** {3 Type definitions} *)
 
-type builtin = ..
-(* Extensible variant type for builtin operations. Encodes in its type
-   arguments the lengths of the expected ty and term arguments respectively. *)
+type builtin = < ty_cst : ty_cst ; term_cst : term_cst > Builtin.t
+(* Extensible variant type for builtin operations. *)
```
bobot, line 22: The type variable in Builtin doesn't seems to be used, is it true?

Gbury: Ok, so about the builtins:
- for the upcoming work on printing terms, it was necessary for the builtins to be in a separate module, and to be parameterized (since the printers will also be parameterized/functorized). 
- however, some of the builtins (mainly the ones about ADT constructors/destructors/testers) carry in their payload some terms, and thus, the builtin type had to be parameterized over all that could appear within.
- in a first try I had explicit type parameters for builtins for each potential type/term construction, but that meant having some 6 type parameters, which was... annoying to use let's say
- thus I used the trick of using a phantom object type as some kind of "type dictionary" of type parameters (each method correspond to a type parameter), which ends up being quite nice, since it brings subtyping of builtins as needed, and can easily be extended.
Though, indeed, now that I re-read this, to allow users to correctly extend builtins, I'll need to add all the usable type parameters in the definition in `expr.ml/expr.mli`, i'll do that asap.

Issue #83, opened 2021-08-02
Comments for suppression


andrewvaughanj:
I'm interested in using Dolmen alongside a test-case reducer such as [`ddSMT`](https://github.com/ddsmt/ddSMT) or [`cvise`](https://github.com/marxin/cvise). However, what I'm *really* interested in is in making sure that the reduction process does not make things *any worse* than when they started.

This ability to suppress warnings (in the input file) would match what typical code-based static analysers allow us to do.

For example, consider a file such as:

```
(set-logic QF_UFLIA)
(declare-fun fmt0 () Int)
(declare-fun fmt1 () Int)
(declare-fun s_count (Int) Int)
(assert (>= (* 4 (s_count (+ (- 2) (* (- 1) fmt0) fmt1))) 0))
(check-sat)
(exit)
```

which gives:

```
File "./suppress.smt2", line 5, character 12-57:
Error This is a non-linear expression according to the smtlib spec.
      Hint: multiplication in strict linear arithmetic expects an integer or
        rational literal and a symbol (variable or constant) but was given:
        - an integer coefficient
        - an arbitrary expression with top symbol not in Arithmetic
```

I would like the ability to make Dolmen *not* report on this error, if I do something like:

```
(set-logic QF_UFLIA)
(declare-fun fmt0 () Int)
(declare-fun fmt1 () Int)
(declare-fun s_count (Int) Int)
; dolmen-no-error
(assert (>= (* 4 (s_count (+ (- 2) (* (- 1) fmt0) fmt1))) 0))
(check-sat)
(exit)
```

Or something similar.

I think to *actually* make this useful, the errors that come from Dolmen need to have some kind of "unique identifier", such that you can silence *particular* warnings (e.g., with gcc can you do things like `-Wno-unused`).

For example, the above might be:

```
; dolmen-no-error[non-linear]
```

or

```
; dolmen-no-error[dol123]
```

where `dol123` is some identifier that Dolmen gives to "This is a non-linear expression according to the smtlib spec".

It is then up to the minimiser to know if it can/can't remove something -- for example, the reducer might simplify Line 5 such that the comment can *then* be removed. It has a weakness that the minimiser might "squash lines together" to get them to avoid errors, but /c'est la vie/.

Gbury:
With the number of warning increasing I've already been thinking about a way to turn some warnings on/off, à la GCC, so it seems like it's the right time to do it, ^^
[...]

PR #84, opened 2021-09-24
Allow users to control severity of warnings

Gbury:
Solves #83 

The main addition of this PR is to add a `-w` option to the `dolmen` binary to allow users to change the severity of a warning. Warnings can thus be at one of three severity:
- disabled, meaning that the warning will not be emitted
- enabled, in which case the warning will simply be emitted/printed
- fatal, which makes the emission of the warning into an error (but still printed as a fatal warning to distinguish it form errors, which cannot be turned off)

To make this possible, each warning (and each error) has an associated mnemonic: a (somewhat) short name for the warning, such as `unused-term-var`. One can then do `dolmen -w=@unused-term-var` to make the warning fatal, or `dolmen -w=-unused-term-var` to disable the warning.

To help with using mnemonics, there are another two new options added to the dolmen cli:
- `dolmen --list` will print a list of all errors and warnings, together with their mnemonic and a short description
```shell-session
$ dolmen --list
mnemonic                      kind           category       description
----------------------------------------------------------------------------------------------------
all                           group          General        All warnings
almost-linear-expr            w:fatal        Typing         Non-linear expression in linear arithmetic
arith-type-expected           error          Typing         Non-arithmetic use of overloaded arithmetic function
arith-type-specific           error          Typing         Incorrect use of overloaded arithmetic function
bad-cstr-arity                error          Typing         Incorrect arity for constructor application
bad-index-arity               error          Typing         Incorrect arity for indexed operator
bad-op-arity                  error          Typing         Incorrect arity for operator application
bad-poly-arity                error          Typing         Incorrect arity for type arguments of a term application
bad-term-arity                error          Typing         Incorrect arity for term application
bad-tptp-kind                 error          Typing         Invalid kind for a TPTP statement
bad-type-arity                error          Typing         Incorrect Arity for type constant application
empty-header-field            w:enabled      Header         Header field with a missing value
...
```

- `dolmen --doc=<mnemonic>` will print a more detailed documentation for the warning/error identified by the mnemonic. Currently, I have not yet written the long documentation of warnings/errors, but that should be done within the next few days.
```shell-session
$ dolmen --doc unused-term-var
                       
  Unused term variable

kind: warning
Category: Typing
Mnemonic: unused-term-var
By default: enabled

  No documentation yet
```

There are a few things yet left to do to complete this PR, but I'd like to gather some feedback while I'm working on the last few things. So here are the things left to do, and some things on which I can use feedback:
- the ability to turn warnings on/off through comments inside input files is not yet done, I'm still thinking about how to do it somewhat properly
- the exact format in which the documentation of warnings is not perfect, if you have any suggestions, I'm all ears
- is the cli interface nice enough for users ? (compared to e.g. gcc's `-Wno-some-foobar` )
- is this enough for your use-case (as described in #83) @andrewvaughanj ?

### DIFF
--- a/src/loop/typer.ml
+++ b/src/loop/typer.ml
@@ -122,6 +122,10 @@ let decl_loc d =
   | Abstract { loc; _ }
   | Inductive { loc; _ } -> loc
 
+let print_bound_kind fmt = function
+  | `Quantified -> Format.fprintf fmt "quantified"
+  | `Letbound -> Format.fprintf fmt "let-bound"
+
 let print_reason fmt r =
   match (r : T.reason) with
   | Builtin ->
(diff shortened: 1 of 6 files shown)


Gbury:
For instance, currently, the `-w` option accepts a comma-separate lists of '[mod]<mnemonic>' where 'mod' is optional and can be one of:
- `'+' to enable the warning corresponding to the mnemonic (which is the default case when no `mod` is given)
- `-` to disable a warning
- `@' to make the warning fatal
- `=' to make the warning enabled but non-fatal (using `+` to "enable" a warning that is already fatal will not do anything)

I expect that these modifiers could be changed to make them a bit nicer to use and/or understand (or even change the cli format for enabling/disabling options, if anyone has a better suggestion).

PR #85, opened 2021-09-28
[Pipeline] Don't print the backtrace directly

bobot:
Give it to the callback along the exception.


Gbury:
It's indeed better to do things that way.
The build failures are because you need to update the `finally` function given to the pipeline in `src/bin/main.ml` and `src/lsp/loop.ml`

PR #87, opened 2021-10-25
Add support for Alt-Ergo's built-ins 

hra687261:


Gbury:
@hra687261  alright, commit f1f05bb281826fa7c808d86fbb21c514d999edad should fix the problem with alt-ergo's implicit type variables in definitions, you'll need to rebase this PR on top of master, ^^

hra687261:
Great, thanks!
I still need to add some tests and the exponentiation operator and whatever might be missing before merging.

Gbury:
> Great, thanks!
> I still need to add some tests and the exponentiation operator and whatever might be missing before merging.

Of course, ^^
I'll also have to do a proper review.

src/languages/ae/lexer.mll
```diff
@@ -117,7 +117,7 @@ let signed_integer = ['-' '+']? integer
 let exp = ['e' 'E'] signed_integer
 let real_exp = digit+ exp
 let decimal = (digit+ '.' digit*) | (digit* '.' digit+)
-let real_dec = decimal exp
+let real_dec = decimal | decimal exp
```
Gbury, line 120:
```suggestion
let real_dec = decimal exp?
```

src/loop/typer.ml
```diff
@@ -703,16 +712,31 @@ let unhandled_ast : (T.env * Dolmen_std.Term.t T.fragment) Report.Error.t =
           print_fragment (env, fragment))
     ~name:"Unhandled AST fragment" ()
 
+let bad_farray_arity =
+  Report.Error.mk ~code ~mnemonic:"bad-farray-arity"
+    ~message:(fun fmt _ ->
+        Format.fprintf fmt "Functional array types in Alt-Ergo expect either one or two type \
+        parameters.")
+    ~hints:[text_hint]
```
Gbury, line 720: it doesn't look like this hint is actually used, so it can probably go away (and the parameter of this error turned into `unit`)

src/loop/typer.ml
```diff
@@ -989,9 +1013,21 @@ module Make(S : State_intf.Typer with type ty_state := ty_state) = struct
...
       S.error ~loc st unhandled_ast (env, fragment)
+    (* Alt-Ergo Functional Array errors *)
+    | Ae_Arrays.Bad_farray_arity ->
+      S.error ~loc st bad_farray_arity ""
+    (* Alt-Ergo Arithmetic errors *)
+    | Ae_Arith.Expected_arith_type ty ->
+      S.error ~loc st expected_arith_type (ty, "")
+    | Ae_Arith.Incompatible_Arith_Types (tya, tyb) ->
+        S.error ~loc st incompatible_arith_types
+        ((tya, tyb), "Alt-Ergo's native language is strongly typed. Only the real \
```
Gbury, line 1024: This comment about "the real exponentiation operator" is confusing, particularly since the operator can operate either on ints or reals. The current wording makes it seem as if the exponentiation always return a real (while taking either ints or reals).

src/standard/expr.ml
```diff
@@ -1855,7 +1867,7 @@ module Term = struct
         with_cache ~cache:(Hashtbl.create 13) (fun (i, j, n) ->
             mk' ~builtin:(Builtin.Bitv_extract (i, j))
               (Format.asprintf "bitv_extract_%d_%d" i j) []
-              [Ty.bitv n] (Ty.bitv (i - j + 1))
+              [Ty.bitv n] (Ty.bitv (j - i + 1))
```
Gbury, line 1870: This looks like quite a breaking change: is the current really incorrect ? (which seems surprising given that there were no errors related to bitv extraction when benchmarking the smtlib), or did you just misunderstand the order of the arguments ? (which is admittedly very easy)

In any case, if this change is needed, it should be in a separate PR, with adequate test cases (particularly smtlib test cases).

src/standard/expr.mli
```diff
@@ -1182,6 +1182,9 @@ module Term : sig
     include Dolmen_intf.Term.Tptp_Tff_Arith_Common with type t := t
     (** Satisfy the common interface for TPTP's arithmetic on integers. *)
 
+    val pow : t -> t -> t
+    (** Exponentiation on integers. *)
+
```
Gbury, line 1187: This (and probably a few lines after this), could be removed by adding an `include` for the `Ae` arith module types (similar to the includes for the smtlib and tptp arith module types just above)

src/standard/expr.mli
```diff
@@ -1224,6 +1230,9 @@ module Term : sig
     include Dolmen_intf.Term.Smtlib_Float_Real with type t := t
     (** Satisfy the real part of the SMTLIB's Float requirements *)
 
+    val pow : t -> t -> t
+    (** Exponentiation on integers. *)
+
```
Gbury, line 1235: Same as for the integers.

src/typecheck/arith.ml
```diff
@@ -2,6 +2,95 @@
...
+        | `Real -> mk_real t
+        | _ -> Type._error env (Ast ast) (Expected_arith_type ty)
+
+      let dispatch2 ?(strict = true) env (mk_int, mk_real) ast a b =
+        let tya = T.ty a in
+        let tyb = T.ty b in
+        match Ty.view tya, Ty.view tyb with
+        | `Int, `Int -> mk_int a b
+        | `Real, `Real -> mk_real a b
+        | (`Real, `Int | `Int, `Real) when not strict -> mk_real a b
```
Gbury, line 32: in this case, you should add explicit type coercions on the ``` `Int``` argument, before giving it to the mk_real function

src/typecheck/arith.ml
```diff
@@ -2,6 +2,95 @@
...
+        | (`Real, `Int | `Int, `Real) ->
+          Type._error env (Ast ast) (Incompatible_Arith_Types (tya, tyb))
+        | _ ->
+          Type._error env (Ast ast) (Expected_arith_type tyb)
```
Gbury, line 36: it's better to use `tya` here I'd say (most people assume a left-to-right order of typechecking)

src/typecheck/arith.ml
```diff
@@ -2,6 +2,95 @@
...
+      | Incompatible_Arith_Types : Type.Ty.t * Type.Ty.t -> Term.t Type.err
+
+      let dispatch1 env (mk_int, mk_real) ast t =
+        let ty = T.ty t in
+        match Ty.view ty with
+        | `Int -> mk_int t
+        | `Real -> mk_real t
+        | _ -> Type._error env (Ast ast) (Expected_arith_type ty)
+
+      let dispatch2 ?(strict = true) env (mk_int, mk_real) ast a b =
```
Gbury, line 26: Looks like the `~strict` argument is actually never used in the rest of the code, or did I miss something ?

src/typecheck/arith.mli
```diff
@@ -1,4 +1,27 @@
...
+      (T : Dolmen.Intf.Term.Ae_Arith with type t := Type.T.t
+                                        and type ty := Type.Ty.t) : sig
+
+    type _ Type.err +=
+      | Expected_arith_type : Type.Ty.t -> Dolmen.Std.Term.t Type.err
+      (** Error raised when an arithmetic type was expected (i.e. either
+          int or real), but another type was found. *)
+      | Incompatible_Arith_Types : Type.Ty.t * Type.Ty.t  -> Dolmen.Std.Term.t Type.err
+      (** Raised when a binary application that expects two values of the same
+          arithmetic type receives two values of different arithmetic types. *)
```
Gbury, line 17: Actually, this error could go away, and the code simplified (see my other comment).

src/typecheck/arith.ml
```diff
@@ -2,6 +2,95 @@
...
+        let tya = T.ty a in
+        let tyb = T.ty b in
+        match Ty.view tya, Ty.view tyb with
```
Gbury, line 29: Assuming the `~strict` parameter is not actually used, then you could delete the `Incompatible arith types` error, and this match could be simplified greatly, by simply pattern matching on the type of the first argument, and applying the corresponding `mk_*` function. In that case, when the second argument has a different (but still arith) type, the error would be caught by the generic mechanism for when an argument does not have the expected type. However, it might result in slightly less precise error messages.

src/typecheck/arrays.ml
```diff
@@ -1,6 +1,43 @@
...
+
+    type _ Type.err +=
+      | Bad_farray_arity : Dolmen.Std.Term.t Type.err
+
+    let parse env s =
+      match s with
+      | Type.Id { name = Simple "farray"; ns = Term } ->
+        `Ty (
+          fun ast args ->
+            Base.ty_app2 (module Type) env s Ty.array ast
```
Gbury, line 22: This code is a bit too convoluted. You should try something more along the lines of what is done in the case of the `minus` sign in the smtlib2 arithmetic (see e.g. https://github.com/Gbury/dolmen/blob/f1f05bb281826fa7c808d86fbb21c514d999edad/src/typecheck/arith.ml#L560-L568 )

src/typecheck/bitv.mli
```diff
@@ -1,5 +1,23 @@
...
+(** Smtlib bitvector builtins *)
+module Ae : sig
+
+  module Tff
+      (Type : Tff_intf.S)
+      (Ty : Dolmen.Intf.Ty.Ae_Bitv with type t := Type.Ty.t)
+      (T : Dolmen.Intf.Term.Ae_Bitv with type t := Type.T.t) : sig
+
+    type _ Type.err +=
+      | Invalid_bin_char : char -> Dolmen.Std.Term.t Type.err
```
Gbury, line 13: Missing doc comments

src/typecheck/core.ml
```diff
@@ -12,12 +12,17 @@ module Ae = struct
...
+    let mk_and a b = T._and [a; b]
+
     let parse env s =
       match s with
 
       (* Types *)
       | Type.Builtin Ast.Bool ->
         `Ty (Base.app0 (module Type) env s Ty.bool)
+      | Type.Builtin Ast.Prop ->
+        `Ty (Base.app0 (module Type) env s Ty.bool)
```
Gbury, line 25: Not that it's too big of a problem, but are there cases where we would typecheck something that alt-ergo wouldn't because here we don't make a distinction between bool and prop ? (In any case, it's probably better to leave this as is, and later on be careful when translating typed terms to alt-ergo's hashconsed terms).

src/typecheck/core.ml
```diff
@@ -29,17 +34,54 @@ module Ae = struct
...
+              let else_ = Type.parse_term env c in
+              T.ite cond then_ else_
+          )
+        )
+
+      (* Equality *)
       | Type.Builtin Ast.Eq ->
         `Term (Base.term_app2 (module Type) env s T.eq)
+      | Type.Builtin Ast.Distinct ->
+        `Term (
```
Gbury, line 66: Why not simply do a `Base.term_app_list` with a `T.distinct` here, as is done for e.e.g smtlib ? (Side note: maybe we need to add a parameter to `term_app_list` to specify a minimum number of arguments ?)

src/typecheck/core.ml
```diff
@@ -29,17 +34,54 @@ module Ae = struct
...
+          fun _ast args ->
+            let args = List.map (Type.parse_term env) args in
+            match args with
+            | [x; y] -> T.neq x y
+            | _ -> T.distinct args
+        )
+
+      (* Ignore the AC symbol *)
+      | Type.Id id when Id.equal id Id.ac_symbol ->
+        `Tags (fun _ast _args -> [])
```
Gbury, line 76: You should not ignore the `ac` specification here, but instead add a new tag for it. See e.g. how it is done for `Zf` (cf https://github.com/Gbury/dolmen/blob/f1f05bb281826fa7c808d86fbb21c514d999edad/src/interface/tag.ml#L42-L49 and https://github.com/Gbury/dolmen/blob/f1f05bb281826fa7c808d86fbb21c514d999edad/src/typecheck/core.ml#L394-L395 )

Gbury (changes requested):
Overall this is good, but there are a few changes required.
Also, you need to run `make test` locally to automatically generate the dune and expect files associated to the tests, and commit them.

Gbury:
Also, since I merged #88 , you'll need to rebase on top of master and re-run `make test`/`make promote` since that PR added things to the warning/error output.

src/loop/typer.ml
hra687261: Done.

src/loop/typer.ml
hra687261: Alt-Ergo's native language has two exponentiation operators ```**``` and ```**.```. ```**``` takes only ints and returns ints while ```**.``` can take either ints or reals or both and returns a real. In the comment I was referring to ```**.```.

src/standard/expr.ml
hra687261: Yes, this was wrong. It's because in Alt-Ergo the extract operator takes the indexes of the extraction from start to end, while in Smt-Lib, it's the opposite. I fixed this and just inverted them in [src/typecheck/bitv.ml#L37](https://github.com/hra687261/dolmen/blob/ae-builtins/src/typecheck/bitv.ml#L37).

src/standard/expr.mli
hra687261: Done.

src/standard/expr.mli
hra687261: Done.

src/typecheck/arith.ml
hra687261: According to Alt-Ergo's doc [doc](https://ocamlpro.github.io/alt-ergo/Input_file_formats/Native/02_types/02_01_builtin.html#built-in-types) we should only do that when the ```**.``` is used. But I was wondering if there is a reason why we would not want to do it for the other operators? aka: whenever one of the two operands is a real the other one (if it's an int) is coerced to a real and the mk_real function is applied.

src/typecheck/arith.ml
hra687261: Done.

src/typecheck/arith.mli
hra687261: Done.

src/typecheck/arith.ml
hra687261: Done.

src/typecheck/arrays.ml
hra687261: I am not sure that I understood what you meant, but I slightly changed it in the new commit.

src/typecheck/bitv.mli
hra687261: Fixed.

src/typecheck/core.ml
hra687261: I don't think so. According to [this](https://github.com/OCamlPro/alt-ergo/blob/next/docs/sphinx_docs/Input_file_formats/Native/01_declaration_of_symbols.md#predicate-propositional-valued-functions) prop and bool used to be two different types but are now the same and behave in the same way. And internally in Alt-Ergo there is only a bool type and no prop while it still supports prop in it's input files.

src/typecheck/core.ml
hra687261: Done. (I think that the check function in ```Base.term_app_list``` should be of the same type as the one in ```Base.term_app_left``` that way the caller can define whatever checks he wants to do on the ast and args)

src/typecheck/core.ml
hra687261: Done.

hra687261:
@Gbury where exactly?

Gbury:
ah, github's interface is annoying: when looking at a single commit's diff you can't properly comment, and you have to go back to the page with the whole changes to the PR to make the comment, which i'll do right now, ^^

src/standard/expr.ml
```diff
@@ -841,7 +842,7 @@ module Ty = struct
   let add_tag (t : t) k v = t.ty_tags <- Tag.add t.ty_tags k v
   let add_tag_opt (t : t) k o = t.ty_tags <- Tag.add_opt t.ty_tags k o
   let add_tag_list (t : t) k l = t.ty_tags <- Tag.add_list t.ty_tags k l
-  
+
```
Gbury, line 845: spurious white line added here

hra687261: That wan't intentional, my editor automatically removes spaces at the ends of lines and that line only had two spaces on it so it removed them. I can put them back if you want :)

Gbury: oh, I didn't see that, of course removing trailing white-spaces is good, ^^

hra687261:
@Gbury Just in case you didn't notice, I applied the changes you requested and replied to the unresolved discussions. Can you review them whenever you have the time? Thanks :)

Gbury:
I'll review it as soon as I have the time, ^^

src/interface/term.ml
```diff
@@ -620,19 +620,143 @@ module type Ae_Base = sig
...
+  (** Exclusive disjunction. *)
+
+  val ite : t -> t -> t -> t
+  (** [ite condition then_t else_t] creates a conditional branch. *)
+
+  val distinct : t list -> t
+  (** Distinct constraints on terms. *)
+
+  val neq : t -> t -> t
+  (** Disequality. *)
```
Gbury, line 648: I think we don't need this one anymore since we always use `distinct`

src/interface/term.ml
```diff
@@ -620,19 +620,143 @@ module type Ae_Base = sig
...
+  end
+
+  module Real : sig
+    include Ae_Arith_Common with type t := t
+
+    val div : t -> t -> t
+    (** Exact division on reals. *)
+
+    val to_real : t -> t
+    (** Conversion from an integer term to a real term. *)
```
Gbury, line 722: This should be in the `Int` module (`Real.to_real` is a bad, confusing and anbiguous name).

src/typecheck/arith.ml
```diff
@@ -2,6 +2,102 @@
...
+          `Term (fun ast args ->
+            Base.term_app2 (module Type) env s
+              (fun a b ->
+                let tya = T.ty a in
+                let a', b' =
+                  match Ty.view tya, Ty.view (T.ty b) with
+                  | `Real, `Real -> a, b
+                  | `Real, `Int -> a, T.Real.to_real b
+                  | `Int, `Real -> T.Real.to_real a, b
+                  | `Int, `Int -> T.Real.to_real a, T.Real.to_real b
```
Gbury, line 77: These should be `T.Int.to_real`.

src/typecheck/arrays.ml
```diff
@@ -1,6 +1,42 @@
...
+
+    let parse env s =
+      match s with
+      | Type.Id { name = Simple "farray"; ns = Term } ->
+        `Ty (
+          fun ast args ->
+            match args with
+            | [_] ->
+              let int_ty = Dolmen_std.Term.builtin Dolmen_std.Term.Int () in
+              Base.ty_app2 (module Type) env s Ty.array ast (int_ty :: args)
```
Gbury, line 25: Since you already check the arity explicitly, you don't need the `Base.ty_app` wrapper here, and can directly return the result of applying `Ty.array` to the correct args. (same for the second case).

src/typecheck/bitv.ml
```diff
@@ -1,6 +1,46 @@
...
+
+      (* Bitvector litterals *)
+      | Type.Id { ns = Value Bitvector; name = Simple name; } ->
+        `Term (Base.app0_ast (module Type) env s (parse_binary env name))
+
+      (* Bitvector operators *)
+      | Type.Builtin Bitv_concat ->
+        `Term (Base.term_app2 (module Type) env s T.concat)
+      | Type.Builtin (Bitv_extract (l, r)) ->
+        `Term (Base.term_app1 (module Type) env s (T.extract r l))
```
Gbury, line 37: it's surprising that there isn't a `get` function on bitvectors...

src/typecheck/bitv.mli
```diff
@@ -1,5 +1,25 @@
...
+      (** Error raised when a character that isn't '0' or '1' is
+          used inside a bitvector string *)
```
Gbury, line 15: You also should add a doc comment for the whole type.

src/typecheck/core.ml
Gbury: Ok, great

src/typecheck/core.ml
```diff
@@ -29,17 +35,48 @@ module Ae = struct
...
+          )
+        )
+
+      (* Equality *)
       | Type.Builtin Ast.Eq ->
         `Term (Base.term_app2 (module Type) env s T.eq)
+      | Type.Builtin Ast.Distinct ->
+        `Term (Base.term_app_list (module Type) env s T.distinct)
+
+      (* Ignore the AC symbol *)
```
Gbury, line 69: This comment needs to be updated since we don't ignore the AC symbol anymore.

Gbury (changes requested):
We're almost there ! There are still a couple of changes required (see comments).

Additionally, I think you haven't added support for typing explicit triggers in input files (which is quite important), so you should probably look into that.

src/typecheck/arith.ml
Gbury: Well, it can be useful to do that, but it can also silently turn a lot of int computation into real computations silently if you don't pay attention, so it's a kind of double-edged sword. In any case, it's up to alt-ergo to define what its language specifies regarding that, ^^

hra687261: Ok

src/typecheck/bitv.ml
hra687261: Yes, but Alt-Ergo only supports the extraction and concatenation operators on bitvectors for now.

src/interface/term.ml
hra687261: Yes, I removed it.

src/interface/term.ml
hra687261: Yes, my bad. It's fixed.

src/typecheck/arith.ml
hra687261: Done.

src/typecheck/arrays.ml
hra687261: Done.

src/typecheck/bitv.mli
hra687261: Done.

src/typecheck/core.ml
hra687261: Done.

hra687261:
Thanks for the review! I applied the requested changes.

Regarding the type-checking of explicit triggers and filters I did a first attempt to add it, but it seems that when the triggers and filters that are added to a quantified formula contain the quantified variables in the formula, those variables are seen as out of scope and a typing error is raised. (see: [this](https://github.com/Gbury/dolmen/pull/87/commits/786b598a58238b0017f4d022e309b6e108cb894a#diff-f1b39685b80ac6f05968417708cbe55f51d86c73e23c6236b199afb24de03411) and [this](https://github.com/Gbury/dolmen/pull/87/commits/786b598a58238b0017f4d022e309b6e108cb894a#diff-c70a27d6932db9bd3c8e6c968145ddd332c13f15d18a6f21478155b7ba8db1f4))

src/typecheck/core.ml
```diff
@@ -9,15 +9,31 @@ module Ae = struct
...
+    let mk_and a b = T._and [a; b]
+
+    let parse_trigger env ast = function
+      | { Ast.term = Ast.App (
+          { Ast.term = Ast.Builtin And; _ }, l
+        ); _} ->
+        List.map (Type.parse_term env) l
+      | _ ->
+        Type._error env (Ast ast)
+          (Type.Expected ("A conjunction of terms", None))
```
Gbury, line 27:
```suggestion
          (Type.Expected ("A multi-trigger (i.e. a list of term patterns)", None))
```

Gbury (changes requested):
One minor change left and we'll be good to go !

src/typecheck/bitv.ml
Gbury: Ok.

src/typecheck/arrays.ml
```diff
@@ -1,6 +1,42 @@
...
+
+    let parse env s =
+      match s with
+      | Type.Id { name = Simple "farray"; ns = Term } ->
+        `Ty (
+          fun ast args ->
+            match args with
+            | [ty] ->
+              let int_ty = Dolmen_std.Term.builtin Dolmen_std.Term.Int () in
+              Ty.array (Type.parse_ty env int_ty) (Type.parse_ty env ty)
```
Gbury, line 25: You shouldn't need to go through the type-checker to type the `int` constant. Instead, you should require the int type in the `Ty` functor argument.

hra687261: Done.

Gbury (approved):

Gbury:
Merged, thanks for the PR, :)

PR #89, opened 2021-11-18
Complete and fix alt-ergo's native language typing

hra687261:
The purpose of this PR is to fix Dolmen's support for Alt-Ergo's native language.
I've added some test files that Alt-Ergo accepts but Dolmen doesn't.


tests/typing/pass/ae/farray/farray_2.ae
```diff
@@ -0,0 +1,7 @@
+
+logic t : int farray
+
+axiom a : forall t : 'a farray.
+   t[0] = 1
+
+goal g30 : t[0] = 1
```
hra687261, line 7: In Alt-Ergo, the type of the universally quantified functional array ```t``` is set to: ```(int, int) farray```.
In dolmen if we remove the quantification: 
```
logic t : 'a farray
axiom a : t[0] = 1
goal g30 : t[0] = 1
```
Then it passes the type-checking and ```t```'s type is set to ```(int, int) farray```. 
I think the same thing should be done with the quantified ```t```.

tests/typing/pass/ae/types/types_2.ae
```diff
@@ -0,0 +1,3 @@
+
+type 'a list
+type ('a, 'b) map
```
hra687261, line 3: This one is probably due to how abstract type declarations are handled in [src/standard/statement.ml](https://github.com/Gbury/dolmen/blob/master/src/standard/statement.ml#L409).
Instead of initializing ```list``` as a parameterized type with a list of type variables as parameters, it is initialized as a function type with the type parameters as types for the functions' arguments.
I think the definition of abstract types should look something like this:
```
type abstract = { id: Id.t; ty: term; vars: term list; loc: location; }
```
instead of: 
```
type abstract = { id: Id.t; ty: term; loc: location; }
```

Gbury: There should be no need to change the type of abstract type definition. If there's any problem it's more likely to be that the wrong abstract type declarations are generated by the parser.

Gbury: I'll take a look

tests/typing/pass/ae/farray/farray_2.ae
Gbury: Yeah, the problem is that alt-ergo handles type variables much the same way as ocaml, and implicitly quantifies over them *if they are still free after typechecking* , i.e. you can type something with a type variable that later unifies with e.g. `int` and it won't be a problem. But in the context of an SMT that puts unnecessary constraints on the type-checker, and it makes the problems much less easy to understand, so unless it's really problematic I think that behaviour should be removed from alt-ergo's native language.

src/typecheck/thf.ml
```diff
@@ -1315,7 +1331,20 @@ module Make
...
   and parse_pattern_var ty env ast s =
-    let v = mk_term_var env (Id.name s) ty in
+    let check_pattern_var_id (name: Dolmen_std.Name.t) =
+      match name with
+      | Simple s
+      | Indexed { basename = s; _ }
+      | Qualified { basename = s; _ } ->
+        let c = String.get s 0 in
+        (* Check if the name starts with an uppercase character. *)
+        if int_of_char c >= 65 && int_of_char c <= 90
```
Gbury, line 1341: This whole check is very very wrong. In the typechecker, you cannot suppose that "constructor=first letter capitalised": it's a property of a language, and the typechecker has to correctly handle different languages.
If I understand correctly, the problem is that without this piece of code, using a not-defined constructor with no arguments in a pattern matching would lead to it being understood as a variable, right ? If so, the answer is more likely to introduce a namespace for constructors, and ensure that symbols with that namespace are not turned into variables. (the other way around is not reliable, as most language cannot statically identifiy what is or is not a variable).

src/standard/path.ml
```diff
@@ -37,3 +37,5 @@ let absolute path name = Absolute { path; name; }
 let rename f = function
   | Local { name; } -> local (f name)
   | Absolute { path; name; } -> absolute path (f name)
+
+let get_name (Local { name; } | Absolute { name; _ }) = name
```
Gbury, line 41: This is a dangerous function, and it very likely should not exist: it makes you lose the path information from absolute paths.

src/typecheck/thf.ml
hra687261: Ok, my bad. I assumed that it was the case (that "constructor=first letter capitalized"). But since it isn't I will just remove it. I don't think it's necessary to add a special namespace for constructors since the call to ```find_bound``` guarantees that if the symbol represents a constructor, it will not be considered as a variable.

src/standard/path.ml
hra687261: Ok, I thought it would be a more straight-forward way to get the base-name of local/global variables in the cases in which the path is always empty (like in Alt-Ergo).

Gbury: Well, it's the case for problems coming from the alt-ergo native language (and most languages currently), but it will change with smtlibv3, where there will be modules and module paths used.

src/typecheck/thf.ml
Gbury: Just as an example, in tptp, variables must all start with an uppercase letter, ^^

hra687261: Oups...

Gbury: I guess you'll see but if you just remove this check, i'm afraid your example with the unbound `C` constructor will now pass probably with a redundant error on the last case, because `C` will be turned into a variable (unless something prevents that).

src/standard/path.ml
hra687261: Ok, I removed it.

src/typecheck/thf.ml
hra687261: The behaviour is the same in Alt-Ergo, ```C``` is considered as a variable and the following cases are considered as redundant. Here's another [example](https://github.com/OCamlPro/alt-ergo/blob/next/non-regression/invalid/adts/useless-patterns-vars.ae).
```
[Warning] File "non-regression/invalid/adts/useless-patterns-vars.ae", line 5, characters 1-58:Typing Error: Pattern-matching contains unreachable cases. These cases are removed: sB | sC
File "non-regression/invalid/adts/useless-patterns-vars.ae", line 4, characters 1-77: I don't know (0.0025) (4 steps) (goal g_invalid)
```
```D``` is considered as a variable. 
But AE just removes the useless cases and responds, not sure how to turn the Error into a warning in Dolmen, I am also not sure if it's better to have it as an error or as a warning.

Comment (Gbury) [reply src/typecheck/thf.ml]: Well, if alt-ergo allows variable to start with an uppercase letter, then indeed, there's not much we can do here.

File: tests/typing/pass/ae/farray/farray_2.ae (hunk 1 above)
Comment (hra687261) [reply tests/typing/pass/ae/farray/farray_2.ae:7]: I see, I came across a few similar cases in the non-regression tests:
[non-regression/valid/arrays/testfile-arrays037.ae](https://github.com/OCamlPro/alt-ergo/blob/next/non-regression/valid/arrays/testfile-arrays037.ae
)
[non-regression/valid/arrays/testfile-arrays035.ae](https://github.com/OCamlPro/alt-ergo/blob/next/non-regression/valid/arrays/testfile-arrays035.ae
)
[non-regression/valid/arrays/testfile-arrays030.ae](https://github.com/OCamlPro/alt-ergo/blob/next/non-regression/valid/arrays/testfile-arrays030.ae
)
[non-regression/invalid/arrays/testfile-arrays001.ae](https://github.com/OCamlPro/alt-ergo/blob/next/non-regression/invalid/arrays/testfile-arrays001.ae
)
And these files were added [a long time ago (2017)](https://github.com/hra687261/alt-ergo-fuzz/commit/c569679c69fecd832de6be119b2a277623d4ef27). Therefore I don't think we can remove that feature.

Comment (Gbury) [reply tests/typing/pass/ae/farray/farray_2.ae:7]: As they say; it's not a bug it's a feature, or the other way around, :p .... i'll work on handling that.

File: tests/typing/pass/ae/types/types_2.ae (hunk 2 above)
Comment (Gbury) [reply tests/typing/pass/ae/types/types_2.ae:3]: I just pushed to your branch a fix for that.

File: src/standard/expr.ml [hunk 5]
```diff
@@ -1569,6 +1649,17 @@ module Term = struct
...
+    let in_interval (b1, b2) = mk'
+        ~name:"in_interval" ~builtin:(Builtin.In_interval (b1, b2))
+        "InInterval" [] [Ty.int; Ty.int; Ty.int] Ty.prop
+
+    let maps_to =
+      let a = Ty.Var.mk "alpha" in
+      let b = Ty.Var.mk "beta" in
+      mk'
+        ~name:"maps_to" ~builtin:Builtin.Maps_to
+        "MapsTo" [] [Ty.of_var a; Ty.of_var b] Ty.prop
```
Gbury, line 1661: This is wrong: the type of `maps_to` is not closed. You should add the `a` and `b` type variables in the first list, so that they are properly bound in the type of the `maps_to` constant.

tests/typing/pass/ae/types/types_2.ae
hra687261: Great, Thanks!

src/standard/expr.ml
hra687261: Yes. Done.

tests/typing/pass/ae/farray/farray_2.ae
Gbury: So I pushed two commits to master to fix this. You should try and rebase, ^^

hra687261: Great! Thanks.

hra687261:
@Gbury I think I will change the pattern-matching errors to warnings (that's how they are treated in Alt-Ergo) if that's fine by you?
Other than that I think it has all it needs to type-check Alt-Ergo's native language now.

PR #90, opened 2022-01-07
[Bitv] add the size of the bitvectors in the builtins

bobot:
It simplifies the matching and handling of the builtins of the bitvectors to have directly the size of the bitvectors.

### DIFF
--- a/src/standard/builtin.ml
+++ b/src/standard/builtin.ml
@@ -78,25 +78,41 @@ type _ t +=
 type _ t +=
   | Bitv of int
   | Bitvec of string
-  | Bitv_concat of int * int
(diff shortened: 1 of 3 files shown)


Gbury:
This looks good, but we should have inline records with named fields at least for all the cases where there's more than one `int` parameter. And then, maybe it'd be better to then have inline records for all of the bitv builtins to have some kind of coherence.

Gbury:
@bobot ping on my comment about inline records, ^^

Gbury:
note: I'll probably try and do a release of dolmen this week-end, if you want this PR to be included in the release, ^^

PR #95, opened 2022-02-16
[WIP] Model verification using Dolmen

Gbury:
This PR adds code to enable `dolmen` to check models for SMTLIBv2 files.

# Using it in practice

```
cvc5 --dump-model test_file.smt2 &> test_model.rsmt2
dune exec -- dolmen --check-model=true -r test_model.rsmt2 test_file.smt2
```

# Adding new theories

One can add new theories to the evaluator by adding a new file in `src/model/` which defines a `builtins` function, and add that function to the list of builtins in file `src/loop/check.ml` around line 70.

Inspiration can be taken from `src/model/array.ml` to define new theories. It would be helpful if people can help by contributing the code for the regular smtlib theories (arith, fp, bv, strings) as PR against the `model` branch (cc @bobot ).

# Adding new tests

New tests should go in dedicated sub-folders of `tests/model`. You only need to add a `<name>.smt2` file and its model in `<name>.rsmt2` and the testing machinery will automatically generate everything needed (you'll also probably need to promote the expected output of the tests using `make promote`).

# TODO

- [x] Better naming for answers/etc...
- [ ] Add module documentation
- [x] proper error messages
- [ ] properly handle push/pop/etc... (stack management)
- [x] add theories:
  + [x] Algebraic datatypes
  + [x] integer/real arithmetic
  + [x] floating-point computations
  + [x] ~~Strings~~ (will be left for later)
  + [x] Bitvectors
- [x] ~~lazy evaluation for performances (e.g. for if-then-elses, etc...)~~ (we'll see later if it's useful)
- [x] better code around pipes/pipeline
- [x] accept solver-defined corner cases (e.g. result of a division by zero)


Gbury:
First, note that the result of the model verification can already be currently obtained by using the return/exit code of Dolmen, as documented in its `--help` page:
- `0` means a correct model
- `4` means a typing error
- `6` means an incorrect model

From what I tried (using instructions at https://github.com/SMT-COMP/postprocessors/tree/master/model-validation-track ), the pysmt-based model validator script has the following output, which I don't think is the best example of what we'd want:
```
model_validator_status=VALID
model_validator_error=none
starexec-result=sat
```

Separately, the rules for the SMTCOMP (see https://smt-comp.github.io/2022/rules.pdf ) seems to define yet another output format, of the form `VALID`/`INVALID`/`UNKNOWN`

Gbury:
Note that I can (and will) add the required output if need, but it'd be nice to be sure about what we want to output exactly.

Gbury:
> There is one problem with StarExec that we cannot save the output from the post-processors, except in key-value pairs. I'm not sure what the limits are but it would be helpful to have a very short message, say about 60 characters that is specific to the problem, for quickly seeing what is going on. Of course, one can always debug the problem by running dolmen on the local machine.

I see, I'll try and implement a very compact output for dolmen to help with that.

Gbury:
> There is one problem with StarExec that we cannot save the output from the post-processors, except in key-value pairs. I'm not sure what the limits are but it would be helpful to have a very short message, say about 60 characters that is specific to the problem, for quickly seeing what is going on. Of course, one can always debug the problem by running dolmen on the local machine.

@jhoenicke If you use the latest master, you can now use the following options: `dolmen --max-warn=0 --report-style=minimal`, and in case of errors in the file being checked, that will result in a very minimal output in one of the following form:
- `E:<error-mnemonic>` for an error
- `F:<warn-mnemonic>` for a fatal warning
- optionally prefixed by a line `W:<num_total_warnings>`

The mnemonics are fairly small, and they should all be under 20/30 chars (and they do not contain spaces). This should hopefully be small enough for you to store in a key-value pair ?

PR #96, opened 2022-03-17
Model add Int, Real, Bitvector

bobot:
I think it would be preferable if model is not a dependency of loop, because model will at the end have possibly a lot of dependencies (Zarith, Floatting point, algebraic number, ...)


Gbury:
I reviewed the code and pushed changes directly to your branch.
Thanks a lot !

PR #97, opened 2022-05-30
Add FP to model

bobot:
Depends on the Farith library: https://github.com/ocaml/opam-repository/pull/21434


src/model/fp.ml
```diff
@@ -0,0 +1,126 @@
...
+  Value.ops
+    ~compare:Mode.compare
+    ~print:Mode.pp
+    ()
+
+let ops =
+  Value.ops
+    ~compare:(fun b b' -> F.compare b b')
+    ~print:(fun fmt b -> Format.fprintf fmt "%a" F.pp b)
+    ()
```
Gbury, line 17: Event if they're just aliases, can you add type definitions for the rounding mode and fp numbers, with a small comment stating that we use the exact same representation as  the `Farith` lib ?

src/model/dune
```diff
@@ -5,13 +5,13 @@
   (libraries
     ; external deps
     ; dolmen standard deps
-    dolmen dolmen.std dolmen.intf zarith
+    dolmen dolmen.std dolmen.intf zarith farith ppx_deriving.runtime
```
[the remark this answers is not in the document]
bobot: It is due to the use of (implicit_dependencies false), a type alias from ppx_deriving.runtime appears in the .cmi of farith. We have ideas to solve that in OCaml+Dune but no time for it.

src/model/fp.ml
Gbury: even a `type t = F.t` would be enough, it's more for readability than anything

bobot: Ok. (:thinking: on another topic, do you think you could make a release of the current master? :kissing_heart:)

Gbury: > Ok. ( on another topic, do you think you could make a release of the current master? )

Sorry about the delay, I've done the release (all of the github side is done, still waiting on the merge into opam-repo, see ocaml/opam-repository#21490 )

Gbury:
I've performed a release of master (including a few fixes), and rebased the `model` branch on it, so you'll also need to rebase this PR (though that shouldn't be too hard).

Gbury:
I've cherry-picked all of the commits into the `model` branch, see commits 94205877c3e6dea2babe242e016119168fb50977 , ae9040dd2ef28d2e598373f3b2a1c089f60c7c38 and d5d352a2ed4007c207689f7099366f556989fdb9

PR #99, opened 2022-07-02
Add support for evaluating `bvsub`

4tXJ7f:
Currently, Dolmen fails to verify models on SMT2 inputs that contain
`bvsub` due to a missing case in `bitv.ml`. This commit adds the missing
case to fix the issue along with a test case.


Gbury:
Thanks a lot ! Sorry for the oversight. Will merge once CI is green.

PR #100, opened 2022-07-05
Add fp.rem, fp.toIntegral fp.to_ubv, fp.to_sbv, fp.min, fp.max

bobot:
I haven't finished to add tests and try on benchmarks, but you can take a first look. I haven't handled undefined values.


Gbury:
> @Gbury I think you proposed to add an option to force the logic to `ALL`, is it right?

Indeed. This would only affect type-checking (since model verification does not use the provided logic), but it may relax the constraints on input problems and allow for easier verification of models in benchmarks. I'll try and make a PR for that in the following days.

Gbury:
(btw, you need to promote some test results, since you fixed some typos in error messages)

Gbury:
ah... I see. Indeed, `(error ...)` is not currently parsed in `.rsmt2` files, but it's relatively easy to add it. I guess that in such case, when trying to verify the model, we'd return a model-related error stating that the answer is an error / not a model (and thus cannot be verified).

Gbury:
@bobot About this PR, it all looks good. If I understand correctly, you still want to add some tests before I merge, right ?

If you want to handle user-defined interpretation of partially-defined symbols (e.g. for `min`/`max` and `fp_to_bv` ), you can take some inspiration from what is done for division by zero in arithmetic:
- the typechecking part : https://github.com/Gbury/dolmen/blob/master/src/typecheck/arith.ml#L742
- the model part : https://github.com/Gbury/dolmen/blob/master/src/model/int.ml#L72-L82

PR #104, opened 2022-08-26
Added tags to predicates to differentiate them from functions in Alt-Ergo's logic

hra687261:


src/interface/stmt.ml
```diff
@@ -195,6 +195,13 @@ module type Logic = sig
       i.e f is a function symbol with arguments [args], and which returns the value
       [body] which is of type [ret]. *)
 
+  val pred_def    : ?loc:location -> id -> term list -> term list -> term -> term -> t
+  (** Symbol definition. [pred_def p vars args ret body] means that "p(args) = (body : bool)",
```
Gbury, line 199: does that mean that `ret` is unused (since it's usually the return type of the symbol).

src/standard/id.mli
```diff
@@ -50,6 +50,9 @@ val name : t -> Name.t
 val ac_symbol : t
 (** Used to denote associative-commutative symbols. *)
 
+val predicate_decl: t
+(** Used to differentiate between functions and predicates. *)
```
Gbury, line 54:
```suggestion
(** Used to differentiate between functions and predicates in alt-ergo. *)
```

src/standard/statement.ml
```diff
@@ -498,6 +498,12 @@ let fun_def ?loc id vars params ret_ty body =
     def ?loc id ~vars ~params ret_ty body
   ]
 
+let pred_def ?loc id vars params ret_ty body =
```
Gbury, line 501: Is `ret_ty` really useful here ? given that predicate should have a return type of `bool`

src/languages/ae/parser.mly
```diff
@@ -546,13 +546,13 @@ decl:
...
       let ret_ty = T.prop ~loc () in
-      S.fun_def ~loc p [] [] ret_ty body }
+      S.pred_def ~loc p [] [] ret_ty body }
 
   | PRED p=raw_named_ident
     LEFTPAR args=separated_list(COMMA, logic_binder) RIGHTPAR EQUAL body=lexpr
     { let loc = L.mk_pos $startpos $endpos in
       let ret_ty = T.prop ~loc () in
-      S.fun_def ~loc p [] args ret_ty body }
+      S.pred_def ~loc p [] args ret_ty body }
```
Gbury, line 555: with the change to a dedicated function, you could remove the `ret_ty` from here and instead generate it in the body of the `pred_def` function.

Gbury (changes requested):
Few minor changes required, once they are fixed, this is good to merge

src/interface/stmt.ml
hra687261: Yes

src/languages/ae/parser.mly
hra687261: Done

src/standard/id.ml
```diff
@@ -80,6 +80,7 @@ let tracked ~track ns path =
 (* ************************************************************************* *)
 
 let ac_symbol = mk Attr "ac"
+let predicate_decl = mk Attr "predicate"
```
Gbury, line 83: maybe `predicate_def` rather than `decl` ? Usually decl means declaration i.e. mainly stating the type of a symbol, whereas `def` means definition and thus associates a term/body to the symbol.

src/typecheck/core.ml
```diff
@@ -92,6 +92,10 @@ module Ae = struct
       | Type.Id { name = Simple "ac"; ns = Attr; }->
         `Tags (fun _ _ -> [Type.Set (Tag.ac, ())])
 
+      (* Predicate declaration *)
```
Gbury, line 95:
```suggestion
      (* Predicate definition *)
```

Gbury (changes requested):
Sorry, last change required: use `def`/`definition` instead of `decl`/`declaration` since we are handling definitions rather than declarations here.

src/standard/id.ml
hra687261: Yes, I agree. It's done

Gbury (approved):

PR #105, opened 2022-09-09
Add failing example for polymorphic pattern in psmt2

bobot:
I can try to write the fix, but I think you were not happy with parsing the pattern in the parser. What is the right way to know in Thf that `as` is not an identifier but mark an ascription?


Gbury:
Hmm.. that's annoying. In that case, I think the best solution would be that the parser generates the same AST for `as` in the term case as in the s-expr case, in order to be coherent (so that means doing the same ``` `As | `NoAs ``` dance as in the term case in the parser). I think it's reasonable at least for psmt2. I'll have to think how to do it for regular smt2 (which i'm now realizing has slightly incorrect rules for the s-expr case, which i'll try and fix).

That makes me think about what is planned for smtlibv3, because it'd really nice if it could define attribute values as terms if that's what they're used for.

Gbury:
So, the problem with recognizing the `as` symbol in the typechecker is that the raw smtlib version of it is broken, or at least, non local, in the sense that `((as f t) args)` is actually `f args : t`, and there is currently no way to propagate such information upwards in the typechecker if I'm not mistaken.

So I think the best solution is to do the following:
- alter the parsing rules for s-expr to correctly accept all reserved tokens, following the declarations in page 24/25 of [the smtlib spec](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2021-05-12.pdf). (Note that currently, the smtlibv2 parser does not accept any reserved word, and the psmt2 one accepts `as` and `par`, ideally, all reserved words should be accepted in both parsers).
- create a new function from untyped ast to untyped ast which 'desugars' the use of raw reserved words, or rather that transforms a parsed s-expr into a parsed term. For instance, that function would rewrite `((as f t) args)` (which would use a symbol named "as" ) into `(f args : t)` (using the colon builtin symbol of the untyped ast), but we could also rewrite other uses of reserved words, such as let bindings, foralls, etc... (unfortunately, it duplicates a bit of the parsing work, but I'm not sure there are better non-hackish solutions).
- call that function on s-expr in the typechecker before typing it.

The trouble with tying to parse specifically as terms the value of the attribute `:pattern` is that, in addition to being really specific, the only way to pattern match on it in the parser in order to parse the payload differently would be to make `:pattern` a distinct keyword, which would further complexify the parser.

Gbury:
> Can you take a look at the branch?

Sure ! Sorry about the delay, I haven't had much time to review the code this weekend, I'll do that (and most certainly merge this PR) this evening.

Gbury:
This looks good, merging.

Gbury:
Yes ! I'l try and do a release this week-end.

PR #108, opened 2022-09-30
Improve safety of keys in the loop state

Gbury:
This PR is made to address @bobot comments about the fragility introduced by the switch to a heterogeneous map for the loop state: since we can now easily forget to initialize a key that is later needed.

This PR fixes the situation by grouping state keys by module with an init function: i.e. each module/pipe that defines new keys should also define an `init` function that takes labelled/optional arguments, and that once called, will make sure all the keys of that module/pipe have a default value (or at least are in a state so that the module will not error out when accessing keys, for instance the `Typer` module defines the `check_model` key (for dependency reasons), but does not set it in its `init` function, however that is because it assumes a default `false` value when the key is not present).

With that pattern in mind, one only has to make sure that in user's code, all modules/pipes that are used have their corresponding `init` function called to create the initial state. That way, any new keys added to a module, would then be added to the signature of the `init` function, and thus result in a type error at the creation point of the initial state, prompting the user to provide an initial value for that new key.

Incidentally, it made me realize that I probably would have forgotten to update the lsp part of the code when I add new keys (which would be really bad considering that the lsp is not tested by the CI), so that's a nice improvement, ^^


Gbury:
We could make it so that for each module/pipe `M`, the `init` function sets a private key in the state, and all entrypoints of the pipe check that that this key is set before proceeding, and if it's not set, then output a specific error message stating that the `init` function has not been called ? That's a runtime check, but right now I don't see a way to check that at compile time.

PR #112, opened 2022-10-14
Expose Term Consts

hra687261:
And created an `Array` sub-module in `Term` and `Term.Const` 
Doc was mostly copy-pasted, I can change/improve it if you have suggestions on how it should be done

This could be useful if you don't want to use smart constructors. In [Colibri2](https://git.frama-c.com/pub/colibrics) for example Dolmen term constants are used to build ground terms that are used by Colibri2's solver and currently the only way to build new terms is by using the smart constructors with variables and then substituting the variables with the semantic values which is a bit of a hassle.

### DIFF
--- a/src/standard/expr.mli
+++ b/src/standard/expr.mli
@@ -854,644 +854,644 @@ module Term : sig
     include Dolmen_intf.Term.Tptp_Thf_Core_Const with type t := t
     (** Satisfy the required interface for the typing of tptp's Thf. *)
 
-    val eqs : int -> ty id
+    val eqs : int -> t
     (** n-ary equality. *)
 
-    val distinct : int -> ty id
+    val distinct : int -> t
     (** n-ary disequality. *)
 
-    val _and : int -> ty id
+    val _and : int -> t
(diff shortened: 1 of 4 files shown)


src/standard/expr.mli
```diff
@@ -853,6 +854,648 @@ module Term : sig
     include Dolmen_intf.Term.Tptp_Thf_Core_Const with type t := t
     (** Satisfy the required interface for the typing of tptp's Thf. *)
 
+    val eqs : int -> ty id
```
Gbury, line 857: I think it'd be better here to use either `t` or `term_const` (`ty id` is rather obscure and it's easy to forgot that it is a term constant and not a type constant), same for all the rest of the file

Gbury (changes requested):
A small request to change the type abbreviation used for the signature of term constants, but otherwise, this is good to merge.

src/standard/expr.mli
hra687261: Yes I agree, it's done.

Gbury (approved):

PR #113, opened 2022-11-18
Better settings for the interactive prompt

Gbury:
Offers more control over the interactive prompt printed by the parser module.

cc @hra687261


hra687261:
Great! Thanks :)

PR #116, opened 2022-12-05
Possibility to add preludes to dolmenls

hra687261:
The purpose of this PR is to make it possible for the language server to recognize custom terms/builtins that are not part of the standard languages.

It adds a command line option `-p|--prelude`  that takes a path to a logic file, the command can be repeated multiple times (should I remove the flag and just make it take n files?). The given files are parsed before creating the state of the server and parsing it's input, the parsed statements from the files are added as a prelude before parsing and type-checking the input that is received by the server.

It would be better if it was possible to parse and typecheck the prelude files before the input of the server, but apparently, using the same typing state for all the files that the server treats leads to the merging of all the environments. Maybe this can be fixed by making it possible to copy a typing state.


Gbury (changes requested):
This is definitely a good feature to have. I suppose this is mainly useful when working on colibri2, which has more builtins that the SMTLIB ?

Note about the implementation though: we should probably keep the implementation simple, at the cost of re-parsing and typing the prelude files. The main reason for that is that ideally, each file should specify its preludes, because you likely wouldn't want any file to be opened with the preludes loaded (for instance, if you were modifying files in the SMTLIB benchmarks, you'd probably prefer not to have the lsp use the preludes).

I guess ultimately, the underlying problem is how to specify that a given `.smt2` file is actually using an extension of the smtlib (and in that case, which extension), cc @bobot for that question. I guess we could try and re-use the `set-option`/`set-info` command for that, something along the lines of
```
(set-option :extension "<extension-name>") ; for builtin extensions
(set-option :prelude "<prelude-filenames>") ; for some sort of "include"
```

src/lsp/server.ml
```diff
@@ -10,10 +10,12 @@ class dolmen_lsp_server =
     (* one env per document *)
     val buffers: (Lsp.Types.DocumentUri.t, State.t) Hashtbl.t = Hashtbl.create 32
 
+    val mutable prelude = []
```
Gbury, line 13: This should be set for each file rather than globally (so, it should be bound in the hashtable just above, which maps files to states).

Even if you want to use some preludes for a whole project/directory, it's probably simpler to make it so that your editor is configured to send the correct settings to the lsp for each file since the lsp server cannot know the exact situation.

src/lsp/server.ml
```diff
@@ -39,6 +41,44 @@ class dolmen_lsp_server =
...
+        begin try
+            let prelude_files =
+              begin match settings with
+                | `Assoc [ "preludes", `List l ] ->
+                  List.fold_left (
+                    fun acc s ->
+                      match s with
+                      | `String f -> f :: acc
+                      | _ ->
+                        failwith (
```
Gbury, line 57: You should not use `failwith`/`Failure` as a jump mechanism (it's better used as a way to abort with a message), use a different exception instead (for instance, by declaring a new one).

src/lsp/server.ml
hra687261: Not sure how to do that since the communication is only done through a defined set of notifications. The client can notify the server that a file was opened, closed or changed. It can also notify it that the settings have changed (regardless of whether or not there are files that are open or not) it just states that the settings have changed (for the workspace or globally) and it sends the new ones.

For the client to be able to send (possibly different) settings for each file then there needs to be file specific settings from the client side (which would be the same as using `include` directives).

I don't think that this is the intended use-case for the preludes, if you define a custom theory that was not yet added to the SMT-LIB or custom builtins and you want the server to process files that use them, you will probably have multiple files and use the same preludes for an entire workspace or at least whole repositories of those files. I don't think you'll have different preludes by file, in that case just use `include` directives.

The current version uses the same preludes for the whole workspace, alternatively the preludes can be applied to specific repositories then when the server receives a file, it can check whether it's in one of those repositories or not before determining whether or not to use the preludes.

Gbury: oh good point... I missed the fact that the settings sent by the editor were global and not tied to a file URI, :/

In that case, having a mutable field for the server seems reasonable.

src/lsp/server.ml
```diff
@@ -39,6 +41,44 @@ class dolmen_lsp_server =
     method on_notif_doc_did_change ~notify_back d _c ~old_content:_old ~new_content =
       self#_on_doc ~notify_back d.uri new_content
 
+    method! on_notification_unhandled
+        ~notify_back (n:Lsp.Client_notification.t) =
+      match n with
+      | Lsp.Client_notification.ChangeConfiguration { settings; } ->
```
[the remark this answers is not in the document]
hra687261: Done.

src/lsp/server.ml
hra687261: Done.

hra687261:
I added a couple of changes:
- `range_of_loc` in some cases the loc is (0,0) and that results in negative ranges which leads to errors:
https://github.com/hra687261/dolmen/blob/d59b592d7a6a432502115548e702ba47eb632f8c/src/lsp/diagnostic.ml#L15-L22
- `preprocess_uri` because for some reason, VSCode adds a prefix to the uris that are received through the client notifications, I don't know if other IDEs add it too:
https://github.com/hra687261/dolmen/blob/d59b592d7a6a432502115548e702ba47eb632f8c/src/lsp/server.ml#L28-L34

Gbury:
- About the locs : instead of your current code, you should rather return `start_range` when `Loc.is_dummy l`

- About the `uri` : I'll try and experiment to see what happens with `uri` on vim, but if I recall correctly, I didn't have any problem with `file://` prefixes. I suppose this means that you use dolmen with VSCode ? if so, could you share your configuration (by opening a PR to add it to https://github.com/Gbury/dolmen/blob/master/doc/lsp.md )

hra687261:
Yes I use VSCode, I'll make a separate PR in which I will add the source code of the vscode extension that I use

Gbury (approved):

Gbury:
Thanks ! I'm now eagerly awaiting the VScode PR, :)

PR #119, opened 2023-01-09
Added the arrow sort to the psmt2 parser

hra687261:
It will be available in SMT-LIB version 3, so I thought it was appropriate to add to psmt2 since isomorphism will also be available in version 3. http://smtlib.cs.uiowa.edu/version3.shtml

### DIFF
--- a/src/languages/smtlib2/poly/parser.mly
+++ b/src/languages/smtlib2/poly/parser.mly
@@ -129,10 +129,8 @@ sort:
       in
       let loc = L.mk_pos $startpos $endpos in T.apply ~loc c args }
   | OPEN ARROW h=sort t=sort+ CLOSE
-    { let loc = L.mk_pos $startpos $endpos in
-      let rev_t = List.rev t in
-      let ret = List.hd rev_t in
-      let args = h :: List.rev rev_t in
+    { let args, ret = Dolmen_std.Misc.split_last (h :: t) in
+      let loc = L.mk_pos $startpos $endpos in
       List.fold_right (T.arrow ~loc) args ret }
 ;
 
--- a/src/languages/smtlib2/poly/parser.mly
+++ b/src/languages/smtlib2/poly/parser.mly
@@ -87,6 +87,7 @@ reserved:
   | SET_INFO { "set-info" }
   | SET_LOGIC { "set-logic" }
   | SET_OPTION { "set-option" }
+  | ARROW { "->" }
 ;
 
 s_expr:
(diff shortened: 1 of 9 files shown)


src/languages/smtlib2/poly/parser.mly
```diff
@@ -127,6 +128,12 @@ sort:
         T.const ~loc (f I.sort)
       in
       let loc = L.mk_pos $startpos $endpos in T.apply ~loc c args }
+  | OPEN ARROW h=sort t=sort+ CLOSE
+    { let loc = L.mk_pos $startpos $endpos in
+      let rev_t = List.rev t in
+      let ret = List.hd rev_t in
+      let args = h :: List.rev rev_t in
```
Gbury, line 135: Two remarks:
- for correctness, you're missing a `List.tl rev_t`
- for code cleanliness (and efficiency), you should actually define a function `split_last` in `src/standard/misc.ml` and call it from here, and in that function, avoid the double `rev`

Gbury (changes requested):
Currently, psmt2 is typed as a first-order language, meaning that any partial application will be rejected by the typechecker currently, and thus the main effect of this PR right now is simply to add another way to declare an uninterpreted function, which is of limited usefulness.

Therefore, it'd be best if you share (and even add in the tests) examples of exactly what you'd want to achieve.

src/languages/smtlib2/poly/parser.mly
hra687261: Done

hra687261:
What I wanted was to support higher order functions, at least partially. In my case I will only have sorts of the form `(a -> b)`. 
Maybe a solution would be to not always flatten arrow types? Like differentiate between `a -> (b -> c)` and `a -> b -> c` to trick the type-checker into thinking that it is isn't a partial application? Or somehow differentiate between  the first arrow and the second arrow in `a -> (b -> c)`?

Gbury:
Note that the typechecker of dolmen is perfectly capable of typing higher-order terms. The only problem is that currently `psmt2` is defined as a language that only support first order terms, and thus the typechecker is configured to reject higher order terms when typing terms coming from `psmt2`.

Technically, it's really easy to allow higher-order terms, but that would: 1) fundamentally change the language (since it would go from first-order to higher-order), and 2) worsen error messages in some cases. This comes back to: what is your use case for this ? Does it requires to change the definition of `psmt2`, or is it more of a `colibri`-specific extension ?

Gbury:
Is this still needed in the end ?

hra687261:
Currently it isn't, I'll close it!

PR #120, opened 2023-01-11
Fix link to the smtlib module in documentation

TDacik:
The link to the module `Dolmen_smtlib` in the documentation (http://gbury.github.io/dolmen/0.7/dolmen/index.html) does not work. I guess this can be solved by suggested fix of module's name.

### DIFF
--- a/src/index.mld
+++ b/src/index.mld
@@ -63,7 +63,7 @@ Individual language parsers are available through the following modules:
 {!modules:
 Dolmen_dimacs
 Dolmen_icnf
-Dolmen_smtlib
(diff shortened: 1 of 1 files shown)


Gbury:
Good catch ! I must have forgot to update the module's name when I renamed it from smtlib to smtlib2.
Thanks for the fix, this will fix the documentation for versions from now on, and I'll see what can be done for older versions. (Fixing the versions hosted on the github web pages is easy, but documentation hosted on ocaml.org will be harder to fix, :/ )

Issue #121, opened 2023-01-12
Recursive definition of identifier in AE language


Halbaroth:
Dolmen 0.8 rejects the above AE file
```
type 'a t = Cons of {elt : 'a; tail : 'a t} | Nil

function length (list : 'a t) : int =
  match list with
  | Nil -> 0
  | Cons(_, l') -> length(l') + 1
end

logic l : int t

goal g: length(l) <= 4
```
with the error:
```
File "git/alt-ergo/tests/issues/531.ae", line 6, character 19-25:
6 |   | Cons(_, l') -> length(l') + 1
                       ^^^^^^
Error Unbound identifier: `length`
```

Gbury:
Closed by #123

PR #122, opened 2023-01-12
assume all funs are recursive in AE

hra687261:
While trying to fix the issue #121 I got the following error:
```
File "./tmp.ae", line 6, character 4-15:
6 |   | Cons(_, l') -> length(l') + 1
        ^^^^^^^^^^^
Error The typechecker inferred the implicit type to provide to an
      application.
      That inference lead to infer that type to contain the variable `w1`
      which is not in the scope of the inferred type.
      The application is located at line 6, character 4-15
      Variable `w1` was bound at <location missing>.
```

### DIFF
--- a/src/languages/ae/ast.ml
+++ b/src/languages/ae/ast.ml
@@ -197,8 +197,8 @@ module type Statement = sig
   val record_type : ?loc:location -> id -> term list -> (id * term) list -> t
   (** Record type definition. *)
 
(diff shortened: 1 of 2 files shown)


Gbury:
I know about that, I'm already debugging this, ^^

hra687261:
Great! I guess you can close this one then, or merge it and open another one when you fix the bug, as you wish :)

PR #125, opened 2023-01-23
A full mode for parsing from raw source

Halbaroth:
This PR aims to add the support for full mode when the source is a raw string.
Basically, I added a new function `parse_raw_lazy` in the functor located in `Transformer.ml`.

The incremental mode is still the default choice to parse raw string.

### DIFF
--- a/src/interface/language.ml
+++ b/src/interface/language.ml
@@ -46,5 +46,9 @@ module type S = sig
       Useful to process input from [stdin], or even large files where it would
       be impractical to parse the entire file before processing it. *)
 
-end
+  val parse_full_input :
+    [ `File of string | `Contents of string * string ] ->
+    file * statement list
+  (** Parse the whole given file or string into a list. *)
 
+end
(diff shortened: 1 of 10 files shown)


src/interface/language.ml
```diff
@@ -34,6 +34,8 @@ module type S = sig
   val parse_file_lazy : string -> file * statement list Lazy.t
   (** Parse the whole given file into a list. *)
 
+  val parse_raw_lazy : filename:string -> string -> file * statement list Lazy.t
```
Gbury, line 37: missing doc comment

Gbury (changes requested):
Apart from the missing doc comment, everything is good. 
Once that's fixed, I'll merge, ^^

src/interface/language.ml
Halbaroth: Done :)

Gbury (approved):

Gbury:
Merged, thanks !

PR #130, opened 2023-02-13
Add attributes on all decls/defs and properly attach them

Gbury:
The main difference is that with this PR, we can now attach attributes to an individual definition (rather than attach it to the group). This is used in the case of predicates to correctly attach the predicate attribute to the individual definition, which should fix a bug with alt-ergo.


Halbaroth:
<3

PR #132, opened 2023-02-20
Do not use octet and 'o' for memory size

hansjoergschurr:
When the selected memory size is printed the formatter would print numbers such as "1Go".  However, the parser only accepts "1G".  This commit fixes this by only printing "1G", "1M", "1k", etc.

Furthermore, it also changes the word "octet" to "byte".  The later one is more common.


Gbury:
That makes sense, thanks for the patch !

PR #137, opened 2023-04-20
Fix the support of the `in_interval` trigger in AE

hra687261:
(cf. OCamlPro/alt-ergo#591)
This PR changes the typing of the `in_interval` trigger in AE, currently it is typed as taking three ints
While in AE's legacy typer, it can apparently take any type:
https://github.com/OCamlPro/alt-ergo/blob/959f56880f1dbb9426aed842c345d00c0a9493df/src/lib/frontend/typechecker.ml#L1252-L1258
I added an example that AE's native language parser and typechecker are supposed to support


src/standard/builtin.mli
```diff
@@ -96,7 +96,7 @@ type _ t +=
 type _ t +=
   | In_interval of bool * bool
   (** [In_interval (b1, b2): Int -> Int -> Int -> Prop]:
```
Gbury, line 98: You should also update the type schema on this line.

Gbury (changes requested):
This seems reasonable, and since the `in_interval` builtin is only used by alt-ergo, this should not break things.

hra687261:
@Gbury running the test results in the following error:
```
File "tests/regression/prxx/fpa-theory-2019-10-08-19h00.ae", line 225, character 43-44:
225 |          float(prec, exp_min, mode, x) in [?, i[,
                                                 ^
Error Unbound identifier: `?`
```
In Alt-Ergo the question mark is used as an ADT tester, but it can also be used when it is found as a bound in the `in_interval` trigger:
https://github.com/OCamlPro/alt-ergo/blob/bf86100658e9c6bc45701726908c3f1efaeaf1ad/src/lib/reasoners/intervals.ml#L1065-L1067

What would you suggest as a clean and correct way to add it?

src/standard/builtin.mli
hra687261: Done

Halbaroth:
Yes, it seems the question mark on a bound name means the bound is infinity. For instance when AE checks a semantic trigger of the form `x in [a, b]`, it runs the below functions:
```ocaml
let match_interval_upper {Sy.sort; is_open; kind; is_lower} i imatch =
  assert (not is_lower);
  match kind, max_bound i with
  | Sy.VarBnd s, _ when is_question_mark s -> imatch (* ? var *)
  | Sy.VarBnd _, Minfty -> assert false
  | Sy.VarBnd s, Pinfty -> new_var imatch s sort
  | Sy.VarBnd s, Strict (v, _) -> new_low_bound imatch s sort v false
  | Sy.VarBnd s, Large  (v, _) -> new_low_bound imatch s sort v is_open

  | Sy.ValBnd _, Minfty -> assert false
  | Sy.ValBnd _, Pinfty -> raise Exit
  | Sy.ValBnd vl, Strict (v, _) ->
    let c = Q.compare v vl in
    if c > 0 then raise Exit;
    imatch

  | Sy.ValBnd vl, Large  (v, _) ->
    let c = Q.compare v vl in
    if c > 0 || c = 0 && is_open then raise Exit;
    imatch
```
and
```ocaml
let match_interval_lower {Sy.sort; is_open; kind; is_lower} i imatch =
  assert (is_lower);
  match kind, min_bound i with
  | Sy.VarBnd s, _ when is_question_mark s -> imatch (* ? var *)
  | Sy.VarBnd _, Pinfty -> assert false
  | Sy.VarBnd s,  Minfty -> new_var imatch s sort
  | Sy.VarBnd s, Strict (v, _) -> new_up_bound imatch s sort v false
  | Sy.VarBnd s, Large  (v, _) -> new_up_bound imatch s sort v is_open

  | Sy.ValBnd _, Minfty -> raise Exit
  | Sy.ValBnd _, Pinfty -> assert false
  | Sy.ValBnd vl, Strict (v, _) ->
    let c = Q.compare v vl in
    if c < 0 then raise Exit;
    imatch
```
Here `i` is a domain containing some term `t` and the first argument is the upper (respectively lower bound) to satisfy in order to trigger the trigger.
As you can see, these functions ignore completely the bound with a question mark. That means these bounds are semantically equivalent to `Pinfty` and `Minfty`.

bclement-ocp:
Since question marks in interval triggers are essentially used as patterns in alt-ergo, maybe the simplest thing would be to treat a question mark in that context as a `Wildcard`?

Doing a two-part translation `in_interval` → ≤ in dolmen then ≤ → back to `in_interval` in AE sounds brittle and prone to somebody forgetting one half of the translation at some point in the future.

Gbury:
In that case, the solution would probably be to add plus and minus infinity builtins and use those for the bounds with a question mark, unless you have another suggestion ?

Gbury:
@Halbaroth I'd rather not use wildcard for this: currently wildcards are very specifically used for type inference, and I'd rather not use them for terms to avoid confusion (the handling of wildcards is already sufficiently complicated, adding places that produce wildcards that have a different meaning would make debugging problems harder than necessary for little gain).

hra687261:
I guess something like this?
Looks like there is a variant of the `?` bound which is `?id`, in this case it is simply treated simply as an uninterpreted symbol by AE and it makes it possible to use the same bound in multiple places.
Apparently AE doesn't complain when you introduce an uninterpreted symbol without declaring it, or maybe it doesn't complain in some specific cases?

hra687261:
Although, if we use infinity we'll have to restrict the types of the bounds to integers or real (which probably the only cases in which we use them anyway). But in that case, I'll have to change it in Alt-Ergo's legacy front-end aswell.

Gbury:
Well, the "interval" terminology mainly makes sense for 1-dimensional types (even more so if you have open intervals), so the notion of infinity could always be extended to such types if/when such types are used/added in the future.

hra687261:
Yes, but infinity only makes for unbounded types, I don't see a reason why a bounded type (like bit-vectors) can't be in an interval, therefore what better terminology should be used for such cases? Something like open-bound? no-bound? unbounded?

hra687261:
Ok, maybe that was a bad example because it depends on how the sign is interpreted in the interval. But in general, assuming you have a bounded type, you know its inhabitants and you know that there is an order relation between them. You'd want the infinity in the case of the upper bound to be the biggest value and the smallest one in the case of the lower bound.
I'm just saying that I'm not sure that `infinity` is the best term since we are just saying that it is unbounded on that side.

Gbury:
And that's why I was suggesting using `<=` (and other comparison operators) instead of "in interval", because in_interval is syntactic sugar that is better expressed using regular comparison as soon as you start using open/unbounded bounds, ^^

bclement-ocp:
The issue is that `in_interval` has semantic behavior attached to it that doesn't apply to regular comparisons, no?

So : `x in_interval [?, ?i[` is handled differently than `x < ?i` by alt-ergo.

Gbury:
Indeed.. Anyway, any solution not involving wildcards would basically be okay I think. The only thing worrying me is @Halbaroth 's mention of alt-ergo's using "uninterpreted symbol without declaring it" (with `?id`).

Gbury:
> (P.S. you tagged the wrong person twice in a row :) )

Oups, my bad.

@hra687261 do you have an example that uses these unbound variables ?

hra687261:
There are a couple in the one I added: https://github.com/hra687261/dolmen/blob/a53dee6ea5c6e94b2643601ae95255fe04c325a0/tests/regression/prxx/fpa-theory-2019-10-08-19h00.ae

Gbury:
Right, that's going to be a significant problem, and I'll have to think about how to handle these (as a general point: non-explicitly-bound variables are not a great design, at least in automated deduction, because it's easy to make mistakes).

bclement-ocp:
AFAIU the implicitly bound variables are scoped to the trigger, but we should check.

We should probably hold off on thinking too much about how to deal with these implicitly scoped variables in Dolmen for now, as it looks like it is only used in two places:

 - To define the case-split for absolute values (which we are planning on removing in favor of an if-then-else approach, barring major regressions)
 - And for oddly specific triggers in `abs_real_interval_1` and `abs_int_interval_1` that instantiate the axiom `0 ≤ abs(x)`, but only when the arithmetic engine is unable to prove so. I think we can make the arithmetic engine know that an absolute value is always positive in the first place instead (and/or keep the axiom with no triggers if we are worried about matching interactions).

If we can eliminate these two use cases (and assuming that Alt-Ergo's internal preludes are the only users of this mostly undocumented feature, which should be relatively safe), we can simply remove the feature from the language, and not have to think about it anymore.

Gbury:
That would be great indeed ! Another alternative we discussed with @hra687261 was to add explicit bindings for these variables in triggers, but if it's not necessary, that simplifies things a lot.

bclement-ocp:
To summarize the conclusions from the AE dev meeting:

 - We will represent `in` using regular comparisons in Dolmen, but place the corresponding expressions at a separate attr (e.g. "sem_triggers" instead of "triggers") so that Alt-Ergo can then distinguish these. If there is a naked `?` the corresponding inequality will not be added.
 - Mapsto (`|->`), `not_a_theory_constant`, `is_theory_constant`, `linear_dependency` (for those that are recognized as first-class concepts by Dolmen) should also be added to the `"sem_triggers"` for consistency
 - For the problem of binding identifiers like `?i`, they should be bound at the closest quantifier. This should be done in the parser, when parsing a quantifier with triggers, all the `?i` used in the triggers should be collected and bound.

Gbury:
I'll take over this PR to implement the changes described by @bclement-ocp

Gbury:
Closing in favor of #162

PR #138, opened 2023-04-20
Fix bug in model/bitv.ml

Gbury:
In case of negative inputs to bvsmod/Bitv_smod, the code would try to make a bitv from a potentially negative integer, resulting in an internal error.


Gbury:
cc @bobot who wrote the code

src/model/bitv.ml
```diff
@@ -146,9 +148,11 @@ let builtins _ (cst : Dolmen.Std.Expr.Term.Const.t) =
...
     op2 ~cst ~size:n (fun a b ->
         let b = sbitv n b in
-        let a = sbitv n a in
-        if Z.equal b Z.zero then  from_bitv n a
-        else extract n (Z.sub a (Z.mul (Z.fdiv a b) b)))
+        if Z.equal b Z.zero then from_bitv n (ubitv n a)
+        else begin
+          let a = sbitv n a in
+          extract n (Z.sub a (Z.mul (Z.fdiv a b) b))
+        end)
```
bobot, line 155: Another possible fix would be to replace `from_bitv n a` by `extract n a`.

bobot (approved):
Thanks for the fix. I'm okay with it, just perhaps this other fix could be simpler.

src/model/bitv.ml
Gbury: yes, and it was my first fix, but then I changed it to the current state to be more in line with the other cases (e.g. `bvsrem`) which use the `ubitv`/`from_bitv`.

PR #140, opened 2023-04-24
Support check_sat statement for Alt-Ergo language

Halbaroth:
This PR adds the keyword `check_sat` for the native language of Alt-Ergo.

Besides, I modified the `Prove` description for statement. Now, the payload of `Prove` can extend the assumptions and the goals of the current sequent.


src/interface/stmt.ml
```diff
@@ -167,9 +166,9 @@ module type Logic = sig
   val assert_     : ?loc:location -> term -> t
   (** Add an assertion to the current set of assertions. *)
 
-  val check_sat   : ?loc:location -> term list -> t
+  val check_sat   : ?loc:location -> ?name:id -> term list -> term list -> t
```
Gbury, line 169: there's no need to take two term lists as you never use the second one in this PR. If we ever need the second list, then we'll add another function.

src/loop/typer_intf.ml
```diff
@@ -198,7 +198,7 @@ module type S = sig
   (** The type of top-level assertion statements *)
 
   type solve = [
-    | `Solve of formula list
+    | `Solve of formula list * formula list
   ]
```
Gbury, line 202: You should add doc to say that the tuple is `(hyps, goals)`

src/model/loop.ml
```diff
@@ -500,9 +500,9 @@ module Make
           st, acc
         )
 
-  let check_solve st ~(file : _ Dolmen_loop.State.file) ~loc l =
+  let check_solve st ~(file : _ Dolmen_loop.State.file) ~loc (hyps, _) =
```
Gbury, line 503: It's wrong to ignore the goals here ! The goals should be evaluated to see if at least one of them is satisfied !

src/standard/statement.ml
```diff
@@ -261,10 +264,13 @@ let rec print_descr fmt = function
...
-  | Prove [] -> Format.fprintf fmt "prove"
-  | Prove l ->
+  | Prove { hyps = []; goals = [] } -> Format.fprintf fmt "prove"
+  | Prove { hyps = []; goals } ->
+      Format.fprintf fmt "@[<hov 2>prove:@ %a@]"
+        (Misc.print_list ~print_sep:Format.fprintf ~sep:" ||@ " ~print:Term.print) goals
+  | Prove { hyps; _ } ->
     Format.fprintf fmt "@[<hov 2>prove-assuming:@ %a@]"
-      (Misc.print_list ~print_sep:Format.fprintf ~sep:" &&@ " ~print:Term.print) l
+      (Misc.print_list ~print_sep:Format.fprintf ~sep:" &&@ " ~print:Term.print) hyps
```
Gbury, line 273: You should not ignore the goals here !

src/standard/statement.ml
```diff
@@ -470,10 +473,10 @@ let p_inccnf ?loc () =
 let clause ?loc l = mk_clause ?loc l
 
 let assumption ?loc l =
-  mk ?loc (Prove l)
+  mk ?loc (Prove { hyps = l; goals = [] })
 
 (* Smtlib wrappers *)
-let check_sat ?loc l = mk ?loc (Prove l)
+let check_sat ?loc ?name hyps goals = mk ?loc ?id:name (Prove {hyps; goals})
```
Gbury, line 479: As said above: you never use the `goals` in this function (or rather you always provide the empty list), so there is no need for this argument. Additionally, since this is specifically a smtlib wrapper, I'm not particularly fond of adding a `?name` argument that would be unused in the smtlib parser, :/

Gbury (changes requested):

src/languages/smtlib2/poly/ast.ml
```diff
@@ -172,7 +172,7 @@ module type Statement = sig
 
   (** Checking for satisfiablity *)
 
-  val check_sat : ?loc:location -> term list -> t
+  val check_sat : ?loc:location -> ?name:id -> term list -> t
```
Gbury, line 175: I'm not fond of adding an effectively unused parameter here. Instead of trying to use the same `check_sat` function for both `smtlib` and `ae`, how about `ae` uses a new function, e.g. `prove_sat` that would take the `.name` argument ?

src/loop/typer.ml
```diff
@@ -1931,9 +1931,10 @@ module Make
     | `Clause l ->
       Format.fprintf fmt "@[<v 2>clause:@ %a@]"
         (Format.pp_print_list Print.formula) l
-    | `Solve l ->
-      Format.fprintf fmt "@[<hov 2>solve-assuming: %a@]"
-        (Format.pp_print_list Print.formula) l
+    | `Solve (hyps, goals) ->
+      Format.fprintf fmt "@[<hov 2>solve: %a assuming: %a@]"
```
Gbury, line 1935:
```suggestion
      Format.fprintf fmt "@[<hov 2>solve: %a@ assuming: %a@]"
```

Gbury (changes requested):

src/languages/smtlib2/poly/ast.ml
Halbaroth: Ok :)

Halbaroth:
Done. I also tested the PR still works with my PR on Alt-Ergo ;)

Gbury (approved):

Issue #144, opened 2023-04-27
Incorrect parse for Alt-Ergo language involving "and" token


bclement-ocp:
Consider the following:

```
$ cat and.ae
predicate P() = true and true

goal g: P()
```

Alt-Ergo's legacy frontend is happy with it:

```
$ alt-ergo --frontend legacy and.ae
File "and.ae", line 3, characters 9-12: Valid (0.0003) (2 steps) (goal g)
```

But the dolmen parser barfs:

```
$ alt-ergo --frontend dolmen and.ae
File "and.ae", line 1, character 25-29:
1 | predicate P() = true and true
                             ^^^^
Error Syntax error (state 363) while reading the reserved word 'true'.
```

My guess is that some conflicts are not resolved in the right way.

PR #147, opened 2023-04-28
Re-enable support for "and" inside function/predicate

bclement-ocp:
In #129 support was added for mutually recursive function and predicate definitions using the "and" keyword.

However, the keyword "and" can also be used to denote the boolean conjunection, but now the conflicts are resolved in such a way that:

```
predicate X = true and true
```

is now a syntax error (upon seeing the [and] token, [predicate X = true] gets reduced into a [predicate_def], and a recursive [function] or [predicate] definition is expected).

This patch fixes the issue by marking the appropriate rules (predicate_def, function_def) as inline, propagating enough context to the conflict that it can be resolved appropriately.

Note that we are still not able to parse horrors such as:

```
predicate X = true and true and predicate Y = true
```

but that is OK because those were never supported by Alt-Ergo.

(as an aside, I don't think it's even possible by using a %right associative token for AND in Menhir)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,9 @@ next
   and correctly attach predicate attribute to individual definitions
   (PR#130)
 
+- Restore support for toplevel "and" in non-recursive predicate/function
+  definitions (PR #147, fixes issue #144)
+
 ### Typing
 
 - Ignore arithmetic restrictions when typing model values. This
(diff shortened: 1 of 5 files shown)


bclement-ocp:
Fixes #144

Gbury (approved):
Nice fix. Just waiting for CI to re-run.

Gbury:
Actually, this probably deserves a note in the `CHANGES.md` file

PR #153, opened 2023-05-11
[Draft] use algebraic number for reals

bobot:
[ocaml-flint](https://github.com/bobot/ocaml-flint/) version 0.3 is needed


src/standard/builtin.ml
```diff
@@ -66,6 +66,7 @@ type _ t +=
   | Int | Integer of string
   | Rat | Rational of string
   | Real | Decimal of string
+  | Root_of_with_order of { coeffs: string list; order: string }
```
Gbury, line 69: Considering that at one point, we might need to also cover other representations, we could rather have:
```
type algebraic =
| Ordered_root of { coeffs : string list; order : string; }

type _ t +=
  | Algebraic of algebraic (* instead of Root_of_with_order *)
```

And rename all occurrences of `root_of_with_order` to something like `algebraic_ordered_root` ?

Gbury (commented):
The addition of the new primitive looks fine (modulo some renaming as per my comment).

Gbury:
Looking at the CI, and I'm wondering: is `calcium` not available on windows ?

Gbury:
yeah, it's not the first time the windows CI has had problems because the repo used by `setup-ocaml` for windows runner is old and not updated regularly... I'll try and see what solutions there are (including manually adding an up-to-date repo as you suggested)

Gbury (approved):
I pushed a few changes, mainly some refactoring and documentation, as well as proper errors instead of the `invalid_arg`. The code is good for merging, but we need to figure out how to make CI runs succeed; the installation of `flint`/`calcium` takes way too long: 6h+ on the windows CI which end up being forcefully shut down before completion. Ideally, the `flint`/`calcium`/`arb` packages could try and make use of depexts to rely on systems binary packages for the big C (or C++ ?) libraries.

Gbury:
https://github.com/ocaml/opam-repository/pull/24509 might solve the CI problems. Will try again once it is merged.

PR #156, opened 2023-06-06
Add convenience State.update and State.update_opt

bclement-ocp:
These wrappers around `Stdlib.Map.S.update` allow updating the state in a more functional way than using a sequence of `get` and `set` calls.


Gbury (approved):
Looks good. Maybe this deserves a change entry ? (if only so that future users can more easily find the first release which included these new functions)

bclement-ocp:
I added a CHANGES entry. Should I also include `@since` markers in the docstrings assuming next version is 0.9?

Gbury (approved):
Waiting for the CI to run and i'll merge

PR #160, opened 2023-06-12
Add support for prelude files

bclement-ocp:
This improves Dolmen's prelude support by allowing to provide preludes as files, rather than as statements.

The prelude files are parsed separately from the main file, and can use a different language from the main file. The `logic_file` is reset to the appropriate file during statement generation to ensure that the typer always uses the right language.

Since a single pipe is now expected to potentially include statements from different languages, the `additional_builtins` in the typer is passed the current language as an additional argument, allowing to define language-specific additional builtins. Since this is already a breaking change, `additional_builtins` is also changed from a reference to a `State` key for consistency with the rest of Dolmen's API.

### DIFF
--- a/doc/type.md
+++ b/doc/type.md
@@ -21,21 +21,22 @@ let state =
   |> State.init
      ~debug:false ~report_style:Regular ~max_warn:max_int
      ~reports:(Dolmen_loop.Report.Conf.mk ~default:Enabled)
-     ~logic_file ~response_file
+     ~response_file
      (* these limits are ignored in this example; to actually enforce
         the limits, one has to use the `run` function from `Dolmen_loop.Pipeline` *)
      ~time_limit:0. ~size_limit:0.
+  |> State.set State.logic_file logic_file
   |> Typer_aux.init
   |> Typer.init ~type_check:true
 
 (* We can add some custom builtin theories to type extensions if we want *)
-let () =
+let state =
   (* We define a dummy builtin theory that does nothing *)
-  let new_builtin _env _symbol = `Not_found in
+  let new_builtins _state _lang _env _symbol = `Not_found in
   (* We add it to the builtins used during typechecking (in addition to
      all the theories that will naturally be used depending on the `set-logic`
      statement) *)
-  Typer_aux.additional_builtins := new_builtin
+  State.set Typer_aux.additional_builtins new_builtins state
 
 (* Now we can iter on all the parsed statements *)
 let () =
--- a/src/bin/options.ml
+++ b/src/bin/options.ml
@@ -34,6 +34,8 @@ let () =
 type cmd =
   | Run of {
       state : Loop.State.t;
+      preludes: Dolmen_loop.Logic.language Loop.State.file list;
+      logic_file : Dolmen_loop.Logic.language Loop.State.file;
     }
   | List_reports of {
       conf : Dolmen_loop.Report.Conf.t;
@@ -353,7 +355,7 @@ let mk_run_state
     () gc gc_opt bt colors
     abort_on_bug
     time_limit size_limit
-    logic_file response_file
+    response_file
     flow_check
     header_check header_licenses
     header_lang_version
@@ -377,7 +379,7 @@ let mk_run_state
   |> Loop.State.init
     ~bt ~debug ~report_style ~reports
     ~max_warn ~time_limit ~size_limit
-    ~logic_file ~response_file
+    ~response_file
   |> Loop.Parser.init
     ~syntax_error_ref
     ~interactive_prompt:Loop.Parser.interactive_prompt_lang
@@ -523,6 +525,16 @@ let response_file =
   in
   Term.(const mk_file $ response_lang $ response_mode $ response)
 
+let mk_preludes =
+  List.map (fun f -> mk_file None None (`File f))
+
+let preludes =
+  let docs = common_section in
+  let preludes =
+    let doc = "Optional prelude file to be loaded before the input file." in
+    Arg.(value & opt_all string [] & info ["p"; "prelude"] ~doc ~docs)
+  in
+  Term.(const mk_preludes $ preludes)
 
 (* State term *)
 (* ************************************************************************* *)
@@ -652,7 +664,7 @@ let state =
         gc $ gc_t $ bt $ colors $
         abort_on_bug $
         time $ size $
-        logic_file $ response_file $
+        response_file $
         flow_check $
         header_check $ header_licenses $
         header_lang_version $
@@ -664,10 +676,10 @@ let state =
 (* ************************************************************************* *)
 
 let cli =
-  let aux state list doc =
+  let aux state preludes logic_file list doc =
     match list, doc with
     | false, None ->
-      `Ok (Run { state; })
+      `Ok (Run { state; logic_file; preludes })
     | false, Some report ->
       let conf = Loop.State.get Loop.State.reports state in
       `Ok (Doc { report; conf; })
@@ -688,5 +700,4 @@ let cli =
     let doc = "The warning or error of which to show the documentation." in
     Arg.(value & opt (some mnemonic_conv) None & info ["doc"] ~doc ~docv:"mnemonic")
   in
-  Term.(ret (const aux $ state $ list $ doc))
-
+  Term.(ret (const aux $ state $ preludes $ logic_file $ list $ doc))
--- a/src/loop/parser.ml
+++ b/src/loop/parser.ml
@@ -60,6 +60,12 @@ let parsing_error =
             p_ref prod lexed expected)
     ~name:"Parsing error" ()
 
+let stdin_prelude =
+  Report.Error.mk ~code:Code.parsing ~mnemonic:"stdin-prelude"
+    ~message:(fun fmt () ->
+      Format.fprintf fmt
+        "Standard input can't be used as a prelude.")
+    ~name:"Prelude error" ()
 
 (* Pipe functor *)
 (* ************************************************************************ *)
@@ -119,25 +125,21 @@ module Make(State : State.S) = struct
       | Some `Incremental -> State.warn ?loc st full_mode_switch lang
       | _ -> st
     in
-    State.set State.logic_file { file with mode = Some `Full; } st
+    st, { file with mode = Some `Full }
 
-  let set_logic_file ?loc st old (file : _ file) =
+  let switch_to_full_mode_if_needed ?loc st old (file : _ file) =
     match file.lang with
     | Some Logic.Alt_ergo -> switch_to_full_mode ?loc "Alt-Ergo" st old file
-    | _ -> State.set State.logic_file file st
+    | _ -> st, file
 
-  let set_logic_file ?loc st (new_file : _ file) =
-    let old_file = State.get State.logic_file st in
-    match old_file.lang with
-    | None -> set_logic_file ?loc st old_file new_file
-    | Some l ->
-      begin match new_file.lang with
-        | None -> set_logic_file ?loc st old_file new_file
-        | Some l' ->
-          if l = l'
-          then set_logic_file ?loc st old_file new_file
-          else State.error ~file:new_file ?loc st input_lang_changed (l', l)
-      end
+  let set_logic_file ?loc st (file : _ file) =
+    let old = State.get State.logic_file st in
+    match old.lang, file.lang with
+    | Some l, Some l' when l <> l' ->
+      State.error ~file ?loc st input_lang_changed (l', l)
+    | _ ->
+      let st, new_file = switch_to_full_mode_if_needed ?loc st old file in
+      State.set State.logic_file new_file st
 
   (* Parsing *)
   (* ************************************************************************ *)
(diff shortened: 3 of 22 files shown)


src/loop/parser.ml
```diff
@@ -190,70 +200,98 @@ module Make(State : State.S) = struct
...
+          | `Stdin -> raise (Stdin_prelude file)
+          | _ ->
+            match infer_lang file with
+            | file ->
+              let st, file = switch_to_full_mode_if_needed st file file in
+              let st, file, g = parse_file st file in
+              st, wrap_append ~file gp (wrap_parser ~file g)
+            | exception Logic.Extension_not_found ext ->
+              State.error st extension_not_found ext, gp
+        ) (st, (fun st -> st, None)) preludes
```
Gbury, line 287: Unless I missed something, this fold will make it so that the files are parsed beginning from the last element of the list of prelude, and parsing the first element of the list last, which is the inverse of what i'd expect.
This transformation might be easier by doing something like the following pseudo-code:
```
preludes |> to_gen |> Gen.flat_map (fun file -> wrap_gen (set_logic_file ..) (parse_file file))
```

src/loop/parser_intf.ml
```diff
@@ -33,11 +33,23 @@ module type S = sig
   *)
 
   val parse_logic :
+    ?preludes:Logic.language State.file list ->
     Dolmen.Std.Statement.t list -> state -> Logic.language State.file ->
     state * (state -> state * Dolmen.Std.Statement.t option)
```
Gbury, line 38: Since we're breaking interfaces for the upcoming release, I wonder if it wouldn't be simpler to change this function to instead take as main argument a list of files (rather than a single file).
Additionally, I'm not sure that the list of statements as prelude is really useful (and in any case, it's trivial for the user/caller to manually prepend a chosen list of statements), so we might remove it, considering how we've seen that it's not simple to use it correctly.

Gbury (changes requested):

src/loop/parser.ml
```diff
@@ -119,12 +125,16 @@ module Make(State : State.S) = struct
...
-  let set_logic_file ?loc st old (file : _ file) =
+  let switch_to_full_mode_if_needed ?loc st old (file : _ file) =
     match file.lang with
     | Some Logic.Alt_ergo -> switch_to_full_mode ?loc "Alt-Ergo" st old file
-    | _ -> State.set State.logic_file file st
+    | _ -> st, file
+
+  let set_logic_file ?loc st old (file : _ file) =
+    let st, file = switch_to_full_mode_if_needed ?loc st old file in
+    State.set State.logic_file file st
```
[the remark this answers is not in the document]
bclement-ocp: I was confused for a moment because I thought you referred to `switch_to_full_mode_if_needed` but yeah, that's correct — I will make it a local function.

src/loop/parser.ml
```diff
@@ -190,70 +200,98 @@ module Make(State : State.S) = struct
...
+      | None ->
+        let l, _, _ = Logic.of_filename filename in
+        { file with lang = Some l }
+      end
+
+  let parse_file st (file : Logic.language file) =
+    match infer_lang file with
+    | file ->
+      (* Parse the input *)
+      match file.source with
```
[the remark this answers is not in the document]
bclement-ocp: :facepalm:

src/loop/parser.ml
```diff
@@ -190,70 +200,98 @@ module Make(State : State.S) = struct
...
+      | `Stdin ->
+        let lang, file_loc, gen, _ = Logic.parse_input
+            ?language:file.lang (`Stdin (Logic.Smtlib2 `Latest))
+        in
+        let file = { file with loc = file_loc; lang = Some lang; } in
+        st, file, gen
+      | `Raw (filename, contents) ->
+        let lang = Option.get file.lang in
```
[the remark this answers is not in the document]
bclement-ocp: Yep, I thought I added one, but apparently not.

src/loop/parser.ml
bclement-ocp: Right, the `parse_file` calls are performed in the reverse order, but the resulting generator should be in the right order.

Using `Gen` directly is probably better, yes, I am just not used to it — will change it to your suggestion

src/loop/parser_intf.ml
bclement-ocp: I think from the user's point of view separating preludes and main file makes for a cleaner interface, even if there is no difference implementation-wise.

Agreed with removing the list of statements as prelude, I did not touch it because I did not need to change the interface of `parse_logic` otherwise (sure adding `?preludes` is technically a breaking change but probably won't actually break code in practice).

src/loop/typer_intf.ml
```diff
@@ -143,17 +143,26 @@ module type Typer_Full = sig
   (** This signature includes the requirements to instantiate the {Pipes.Make:
       functor*)
 
+  val init :
+    ?ty_state:ty_state ->
+    ?smtlib2_forced_logic:string option ->
+    ?additional_builtins:(lang -> builtin_symbols) ->
+    state -> state
+
+  val additional_builtins : (lang -> builtin_symbols) key
```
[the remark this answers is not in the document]
bclement-ocp: My first draft actually passed `state` instead of `lang` here, but I was not sure about you wanting to allow the builtins to depend on the file being parsed, it feels hackish.

I am fine with adding `state` there. Do you prefer the type be `state -> lang -> builtin_symbols` or just `state -> builtin_symbols`?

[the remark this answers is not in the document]
Gbury: Let's give both the `state` and the `lang`.

Note that users can already insert code in a pipeline that would match on anything available in the state, and change things (either the `additional_builtins` key, or ref) depending on things that can be more or less reasonable. For instance, users could implement some kind of extensions systems for a language where files would have additional conventions (e.g. `some_file_name.extensions.smt2`) to indicate to add some typing extensions, or some users might change some behavior depending on whether we will later verify a model corresponding to the file (which is actually something that dolmen already does). Anyway, we cannot prevent users from shooting themselves in the foot, and there are legitimate reasons to lookup things into the state in order to decide how to type some file (plus it can already be done currently, it's just a bit harder), ^^

src/loop/parser.ml
bclement-ocp: Actually re-reading the code I think it is correct. `parse_file` doesn't actually parse, it returns a lazy generator (unless I missed something, even in the `Full` case, it is a `S.t list Lazy.t`). `Gen.flat_map` doesn't work because the outer state needs to be threaded through. With `List.fold_left` the outer state is propagated from beginning to end and the included list is built in the right order (note `wrap_append ~file gp (wrap_parser ~file g)` and not `wrap_append ~file (wrap_parser ~file g) gp` which would actually be incorrect due to the definition of `wrap_append`).

(As a side note, `wrap_append` is really a poor name, because it `appends` and then changes the file; open to suggestions)

src/loop/typer_intf.ml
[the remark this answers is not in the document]
bclement-ocp: Oh I know, I may use it in Alt-Ergo to restrict bulitins to the builtin preludes. And yeah, good point about being able to do it from the pipeline, let's pass in the `state`.

src/loop/parser_intf.ml
bclement-ocp: I kept `?preludes:Logic.language State.file list -> state -> Logic.language State.file -> ...` so as not to lose the ability to do what you describe in the future.

While I am at it — should I also change `parse_response`, at least for consistency?

bclement-ocp: One difference is that only the main file is allowed to be `Stdin` actually

src/loop/parser.ml
Gbury: Alright, so after re-reading it, the code is indeed correct, but it's a bit of a headache to make sure.

I don't see the problem with `Gen.flat_map` : it's not really necessary to thread, or even have an "outer" state: it's perfectly fine to change the state only when we begin to parse it (and change it on the fly during the pipeline, instead of at the beginning).

bclement-ocp: Yeah, if we don't need to thread the outer state, then `flat_map` is easier for sure (and we probably can get rid of `wrap_append` entirely). Will change

src/loop/parser_intf.ml
bclement-ocp: > While I am at it — should I also change `parse_response`, at least for consistency?

I have now read the code of `parse_response` and now think that it is fine to leave it as is (:

src/loop/parser.ml
Gbury: Great ! (just for the reference: the "outer" state is mainly here for historical reasons: at the beginning the generators returned for the parsed statements used to be a simple `unit -> parsed_statement` function, and didn't thread the state around. I later fixed that to pass the state around, but left a lot of code that actually updated the outer state)

src/loop/parser.ml
```diff
@@ -190,70 +192,131 @@ module Make(State : State.S) = struct
       let st = State.error st ~file ~loc:{ file = file.loc; loc; } parsing_error (syntax_error_ref, perr) in
       st, None
 
-  let parse_logic prelude st (file : Logic.language file) =
+  let[@inline] infer_lang source (file : Logic.language file) =
+    match source with
+    | `Raw (filename, _) | `File filename ->
+      (* Auto-detect input format *)
```
Gbury, line 198: maybe move that comment inside the `None` case, where we actually do the auto-detection ?

src/loop/parser.ml
```diff
@@ -190,70 +192,131 @@ module Make(State : State.S) = struct
...
+        let file = { file with loc = file_loc; lang = Some lang; } in
+        st, file, gen_finally gen cl
+      | Some `Full ->
+        let lang, file_loc, l = Logic.parse_raw_lazy ~language:lang
+          ~filename contents
+        in
+        let file = { file with loc = file_loc; lang = Some lang; } in
+        (* We are called from inside a [gen'] so the returned generator is
+        always evaluated immediately, there is no need to add laziness *)
+        st, file, Gen.of_list (Lazy.force l)
```
Gbury, line 231: That's not good: the lazy part was actually added so that any lexer/parser exception (raise in case of errors during lexing/parsing), are actually raised by the generator, instead of during the `parse_*` functions (this is actually important because the generator is wrapped so that exceptions are properly watched and transformed into errors acting on the most up-to-date state).

src/loop/parser.ml
```diff
@@ -190,70 +192,131 @@ module Make(State : State.S) = struct
...
+
+  (* [map_st_gen f g] calls [f] on all the values produced by [g] in order,
+     threading the state.
+
+     [f] is *not* a generator but a function in the state monad. *)
+  let map_st_gen f (g : _ Gen.t) : _ gen' =
+    make @@ fun st ->
+    begin match g () with
+    | None -> st, None
+    | Some x -> f x st
```
Gbury, line 290: Is there a reason to put `x` first here ? Intuitively I would rather have `st` first, and then `x`

src/loop/parser.ml
```diff
@@ -190,70 +192,131 @@ module Make(State : State.S) = struct
...
+            [switch_to_full_mode_if_needed] *)
+        match infer_lang source file with
+        | file, lang ->
+          fun st ->
+          let st, file = switch_to_full_mode_if_needed st file file in
+          let st, file, g = parse_file st source file lang in
+          let st = State.set State.logic_file file st in
+          st, Some (make @@ wrap_parser ~file g)
+        | exception Logic.Extension_not_found ext ->
+          fun st -> State.error st extension_not_found ext, None
```
Gbury, line 318: I don't see any reason to split this function as is currently done: we would have the same results if the whole function took both arguments at the begining (and the perf would most certainly be better too), unless I'm missing something.

src/loop/parser.ml
```diff
@@ -190,70 +192,131 @@ module Make(State : State.S) = struct
...
+      in
+      let file = { file with lang = Some lang; } in
+      st, file, (Gen.singleton s')
+
+  (* This is a ['a Gen.t] with a threaded state. *)
+  type 'a gen' = {g: State.t -> State.t * 'a option} [@@unboxed]
+
+  let make g = {g}
+
+  let run {g} = g
```
Gbury, line 255: are the `make` and `run` functions really necessary here ?

Gbury (changes requested):
I think we're almost there: I left a few comments, but aside from that I think we're missing some kind of tests to check whether this is working as intended. The simplest would probably be to add a `--prelude` option to the dolmen cli and use it in a test, for instance to test a combination of the ae fpa prelude and a smtlib2 file ?

src/loop/parser.ml
bclement-ocp: Not really, I just don't like "exposing" `{g}`. I can remove them.

Gbury:
You might need to rebase on `master` to fix the CI runs (a commit I pushed earlier contained an obscure bug related to newlines in error messages on windows).

src/loop/parser.ml
bclement-ocp: The way this function is split is simply a byproduct of the iterative refactoring that was performed, there is no other reason. I will make that change

src/loop/parser.ml
bclement-ocp: I tried to keep the `state -> state * 'a` intact because I think of it as a single type constructor; I can flip the arguments around.

src/loop/parser.ml
bclement-ocp: Actually the whole function can't take both arguments because the split is to eliminate `` `Stdin`` from the variant type, but `infer_lang` is only used here now so I will inline it.

bclement-ocp:
@Gbury This should be good to go now

doc/type.md
```diff
@@ -21,21 +21,21 @@ let state =
   |> State.init
      ~debug:false ~report_style:Regular ~max_warn:max_int
      ~reports:(Dolmen_loop.Report.Conf.mk ~default:Enabled)
-     ~logic_file ~response_file
+     ~response_file
```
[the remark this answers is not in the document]
bclement-ocp: Ah I did in `tuto.md` but forgot `type.md`

src/bin/options.ml
```diff
@@ -523,6 +525,16 @@ let response_file =
   in
   Term.(const mk_file $ response_lang $ response_mode $ response)
 
+let mk_preludes =
+  List.map (fun f -> mk_file None (Some `Full) (`File f))
```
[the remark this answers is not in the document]
bclement-ocp: Because I did not think when replicating Alt-Ergo's behavior :) Fixed it

src/loop/parser.ml
```diff
@@ -60,6 +60,12 @@ let parsing_error =
             p_ref prod lexed expected)
     ~name:"Parsing error" ()
 
+let stdin_prelude =
+  Report.Error.mk ~code:Code.generic ~mnemonic:"stdin-prelude"
```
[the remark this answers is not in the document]
bclement-ocp: No reason

Gbury (approved):

PR #162, opened 2023-06-14
Semantic triggers

Halbaroth:
This PR replaces in the typechecker the operator `in_interval` by inequalities and binds variables introduced by the symbol `?`.


src/typecheck/core.ml
```diff
@@ -116,9 +117,61 @@ module Ae = struct
...
+              in
+              let res =
+                match var1 with
+                | Some v -> Ast.exists ~loc [v] res
+                | None -> res
+              in
+              match var2 with
+              | Some v -> Ast.exists [v] res |> Type.parse_term env
+              | None -> Type.parse_term env res
+          )
```
bclement-ocp, line 174: I think there needs to be a wrapper constructor here to indicate that we are in a semantic trigger

src/interface/term.ml
```diff
@@ -305,7 +305,7 @@ module type Logic = sig
 
   (** {3 Triggers} *)
 
-  val in_interval : ?loc:location -> t -> (t * bool) -> (t * bool) -> t
+  val in_interval : ?loc:location -> t -> t * bool -> t * bool -> t
```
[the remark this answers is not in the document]
bclement-ocp: Ah right yes it's parsed then erased at typechecking, my bad

src/standard/expr.ml
```diff
@@ -1896,8 +1896,8 @@ module Term = struct
         [a; b] [Ty.of_var a] (Ty.of_var b)
 
     let in_interval (b1, b2) = mk'
-        ~name:"in_interval" ~builtin:(Builtin.In_interval (b1, b2))
-        "InInterval" [] [Ty.int; Ty.int; Ty.int] Ty.prop
+      ~name:"in_interval" ~builtin:(Builtin.In_interval (b1, b2))
+      "InInterval" [] [Ty.int; Ty.int; Ty.int] Ty.prop
```
bclement-ocp, line 1900: I believe this needs to be α → α → α, not `int → int → int` (at least it can also be `real → real → real`)

src/typecheck/core.ml
```diff
@@ -116,9 +117,61 @@ module Ae = struct
...
+                    let i2 = inequality ~strict:b2 ~lower:false ~loc t2 t in
+                    Ast.and_ ~loc [i1; i2]
+              in
+              let res =
+                match var1 with
+                | Some v -> Ast.exists ~loc [v] res
+                | None -> res
+              in
+              match var2 with
+              | Some v -> Ast.exists [v] res |> Type.parse_term env
```
bclement-ocp, line 172: Is this binding the variable at the right level? I think it needs to be at the level of all the triggers (I think the constructor is `Triggers`, not at the level of the `in` — otherwise we forget that we are using the same variable if it occurs in multiple triggers

src/standard/expr.ml
Halbaroth: Sure! I completely forgot it :)

src/interface/term.ml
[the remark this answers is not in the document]
Halbaroth: In my first attempt, I did a part of the modification in the Parser but it is not simpler.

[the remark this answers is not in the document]
bclement-ocp: No I think this is right, we don't want to do this in the parser, I had a brain fart.

(Actually I believe @Gbury suggested doing this immediately after parsing but before typing, I don't remember the details but I'm sure he will :) )

src/typecheck/core.ml
Halbaroth: I think you are right. Since there is no explicit bindings in the semantic triggers for these variables, I guess it means the formula is handled as a prenex form and all the quantifiers are at the top level. Nevertheless, we should check if it is the case.

bclement-ocp: The quantifiers must be bound at the level of the closest trigger group (`[ ... ]` brackets).

src/interface/term.ml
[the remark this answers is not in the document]
Halbaroth: He told me to do it in the typechecker ;)

Gbury:
I pushed changes to this PR. I'll re-read all the changes tomorrow and merge it, but at least the API should be stable.

src/standard/builtin.mli
```diff
@@ -94,23 +94,24 @@ type _ t +=
...
-  (** [Maps_to: 'term_var -> 'term -> 'term]:
+  (** [Maps_to: 'term_var -> 'term -> Prop]:
       Used in semantic triggers for floating point arithmetic.
       See [alt-ergo/src/preludes/fpa-theory-2017-01-04-16h00.ae].
 
       [warning:] It is an Alt-Ergo semantic trigger that should only be
       allowed inside theories. *)
 
+  | Semantic_trigger
+  (** [In_interval: 'a. 'a -> Prop]:
```
bclement-ocp, line 111:
```suggestion
  (** [Semantic_trigger: 'a. 'a -> Prop]:
```

src/standard/term.mli
```diff
@@ -123,6 +123,8 @@ type builtin =
   | Record_access
   (** Record field access *)
 
+  | Multi_trigger
+  (* Mutli-triggers *)
```
bclement-ocp, line 127:
```suggestion
  (* Multi-triggers *)
```

src/typecheck/arith.ml
```diff
@@ -38,6 +38,37 @@ module Ae = struct
...
+          (parse_bound env lower) (Some t)
+      in
+      let upper =
+        Misc.Options.map2 (mk strict_upper)
+          (Some t) (parse_bound env upper)
+      in
+      match lower, upper with
+      | None, None -> Type.T._true
+      | None, Some b
+      | Some b, None -> b
```
bclement-ocp, line 68:
```suggestion
      | None, Some b
      | Some b, None -> T.semantic_trigger b
```

I believe?

src/typecheck/core.ml
```diff
@@ -18,14 +18,48 @@ module Ae = struct
...
+      let l = List.map Ast.const (free_qm_ids ast) in
+      let t = Ast.exists l ast in
+      let var_infer = Type.var_infer env in
+      let env =
+        Type.with_var_infer env (
+          { var_infer with
+            infer_term_vars_in_binding_pos =
+              Wildcard (Any_base {
+                  allowed = [ Ty.int; Ty.real ];
+                  preferred = Ty.real;
```
bclement-ocp, line 42: Nice!

Gbury (approved):

src/interface/tag.ml
```diff
@@ -34,8 +34,9 @@ module type Ae_Base = sig
   val named : string t
   (** A tag used to name formulas in alt-ergo. *)
 
-  val triggers : term list list t
-  (** Multi-triggers that can be added to quantified formulas *)
+  val triggers : term list t
+  (** Multi-triggers that can be added to quantified formulas.
+      Each term is a mutli-trigger represented as a confjunction of triggers. *)
```
bclement-ocp, line 39:
```suggestion
      Each term is a multi-trigger represented as a conjunction of triggers. *)
```

src/interface/tag.ml
```diff
@@ -55,9 +56,11 @@ module type Smtlib_Base = sig
     (** A tag used to named formulas in smtlib.
         Should correspond to the `:named` attribute. *)
 
-    val triggers : term list list t
+    val triggers : term list t
     (** Multi-triggers (typically annotated on the body of
-        a quantified formula and not the quantified formula itself). *)
+        a quantified formula and not the quantified formula itself).
+        Each term is a mutli-trigger represented as a
+        confjunction of triggers. *)
```
bclement-ocp, line 63:
```suggestion
        Each term is a multi-trigger represented as a
        conjunction of triggers. *)
```

src/interface/term.ml
```diff
@@ -681,7 +683,10 @@ module type Ae_Base = sig
   val distinct : t list -> t
   (** Distinct constraints on terms. *)
 
-  val in_interval : t -> bool * bool -> t -> t -> t
+  val multi_trigger : t list -> t
+  (** Create a multi trigger from a list of arbtirary terms. *)
+
+  val semantic_trigger : t -> t
   (** Semantic trigger: "in interval" check. *)
```
bclement-ocp, line 690: This should either be renamed back to `"in interval"` or the documentation updated to indicate that this is a marker term for semantic triggers (which may or may not be "in interval" checks).

src/interface/term.ml
```diff
@@ -741,8 +746,8 @@ module type Ae_Arith = sig
   val real : string -> t
   (** Real literals *)
 
-  val _and : t list -> t
-  (** Conjunction of formulas *)
+  val semantic_trigger : t -> t
+  (** Semantic trigger: "in interval" check. *)
```
bclement-ocp, line 750: Same as above

src/typecheck/arith.ml
```diff
@@ -11,32 +11,69 @@ module Ae = struct
...
+      | Symbol { ns = Term; name = Simple "?" } -> None
+      | _ ->
+        let t = Type.parse_term env ast in
+        Some t
+
+    let parse_in_interval env ~strict_lower ~strict_upper _ (ast, lower, upper) =
+      let t = Type.parse_term env ast in
+      let t_ty = T.ty t in
+      let x = Type.T.Var.mk "x" t_ty in
+      let x_t = Type.T.of_var x in
```
bclement-ocp, line 52: Is this not guaranteed to be `t_ty`?

src/typecheck/core.ml
```diff
@@ -114,33 +152,17 @@ module Ae = struct
             [Type.Set (Tag.filters, l)]
           )
 
-      (* Semantic triggers *)
-      | Type.Builtin (Ast.In_interval (b1, b2)) ->
-        Type.builtin_term (Base.term_app3_ast (module Type) env s
-                 (fun _ast _a1 _a2 _a3 ->
-                    T.in_interval _a1 (b1, b2) _a2 _a3))
+      (* Mutli-triggers *)
```
bclement-ocp, line 155:
```suggestion
      (* Multi-triggers *)
```

bclement-ocp (approved):

src/typecheck/arith.ml
Gbury: Ah no, it's a bit obscured by all the functors, but in this case, `Type` is actually the typechecker, and therefore `Type.T` is the term module that was given to instantiate the typechecker functor, so `x_t` is actually the term (and not type) that simply represents the variable `x`.

PR #165, opened 2023-06-16
Propagate attributes from `Pack` Statements

Gbury:
These should help recover the structure of `Theory extends...` from alt-ergo files.


bclement-ocp (approved):
This looks sensible and does allow to get the theory information in Alt-Ergo.

PR #166, opened 2023-06-17
Don't load all preludes if one of them is opened

hra687261:
Makes it possible to open prelude files without the typer complaining about multiple definitions. (and load the other prelude files which were defined before the one that was opened)

### DIFF
--- a/src/lsp/server.ml
+++ b/src/lsp/server.ml
@@ -34,7 +34,7 @@ let preprocess_uri uri =
   else uri
 
 let mk_prelude files =
-  List.map (
+  List.rev_map (
     fun f ->
       let dir, file = Dolmen_loop.State.split_input (`File f) in
       Dolmen_loop.State.mk_file dir file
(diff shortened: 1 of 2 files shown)


src/lsp/server.ml
```diff
@@ -34,7 +34,7 @@ let preprocess_uri uri =
   else uri
 
 let mk_prelude files =
-  List.map (
+  List.rev_map (
```
Gbury, line 37: why the `rev_map` here ?

hra687261: Because a `fold_left` is used in the `parse_settings` function, so the list of files is reversed, then re-reversed in `mk_prelude` to get back its order.

Gbury: Hm... in this case the lists will always be "short", so I'd rather we always keep the lists in the correct order even if that means less efficient traversals or a few more `List.rev`.

hra687261: True, I'll change it

hra687261: Done

Gbury:
Note: you'll need to rebase on `master` in order to make CI pass.

Gbury (approved):

PR #168, opened 2023-06-26
Fix bvsdiv and fp.to_ubv fp.to_sbv

bobot:


Gbury (approved):
This looks good. While we're at it, could you try and fix the documentation in `builtin.mli` (I noticed there's a lot of `Fp(s,e)` which should probably be `Fp(e,s)`).

Gbury:
Last question before merging: is it specified in the smtlib that `bvsdiv` returns `-1` or `1` depending on the sign of the first argument when the second argument is `0` ? or is it unspecified ? or a third option ?

Gbury:
Indeed, I was that it was defined as syntactic sugar for a more complex expressions, but I didn't have the motivation to try and understand what the complex expression was doing. In any case, I'll trust you on the specification, and the current code is quite simple so it's good.

PR #181, opened 2023-08-26
Run `realpath` on the paths of LSP prelude files

hra687261:
To make all paths absolute, the server can't find them otherwise.

### DIFF
--- a/src/lsp/server.ml
+++ b/src/lsp/server.ml
@@ -11,7 +11,7 @@ let parse_settings settings =
     List.map (
       fun s ->
         match s with
-        | `String f -> f
+        | `String f -> Unix.realpath f
         | _ ->
           raise (ServerError (
               Format.asprintf


Gbury:
That seems reasonable, however `Unix.realpath` only exists on ocaml `>= 4.13` (as seen by the CI that fails on earlier version of the compiler). It's not clear what could be the solution.

hra687261:
Right, that's unfortunate, there doesn't seem to be any other stdlib alternative, Core offers a realpath function but I don't think we want to add a dependency just for that, or call the C realpath function, or open a process ... I think I'll just add a minimal one that replaces `~`, environment variables, `.` and `..`.

bclement-ocp:
Is `realpath` actually needed? Maybe simply prepending relative paths to the current directory would suffice:

```ocaml
let f = if Filename.is_relative f then Filename.concat (Sys.getcwd ()) f else f in
```

It doesn't handle `~` or environment variables, but then `Unix.realpath` doesn't either.

Gbury:
@hra687261 did you try @bclement-ocp 's suggestion ?

hra687261:
Sorry, forgot about this one.
The lsp server is run on the working directory, so giving it a relative path to the working directory works without appending the relative path to the absolute path of the working directory.
The text of the PR is wrong, I must've tried to use a relative path to the home directory. Which was my initial goal as it would allow to keep common configs in one place instead of duplicating them over work-spaces, but it's fine, I suppose we can keep it this way for now.

PR #182, opened 2023-08-27
Add the String theory to the `ALL` logic of SMT-LIB

hra687261:
The added test didn't pass before, not sure if there are other necessary changes, but this seems to fix the issue.


Gbury:
The fix is good indeed, but a more accurate description would be: "add the String theory to the `ALL` logic of smtlib". I'm not sure the test is needed (it's unlikely for the string theory to disappear and cause a regression).

hra687261:
Alright, removed it.

Gbury (approved):

PR #198, opened 2023-10-31
Quoted symbols and printing for smtlib-related languages

### LINKED ISSUE #196: Quoted identifiers can be keywords
According to my reading of 3.2

Gbury:
This PR does a few things:
- it adds printing functions for identifiers for `smtlib2.6` and `psmt2`
- it accepts quoted reserved words as symbols instead of reserved words, allowing declarations such as
```smt2
(declare-fun |assert| () Bool)
```

Will close #196


src/languages/smtlib2/poly/lexer.mll
```diff
@@ -134,10 +134,19 @@
       | '\n' -> newline lexbuf
       | _ -> ()
     done;
-    (* Check whetehr the symbol is a reserved word. *)
+    (* Check whether the symbol is a reserved word. *)
     try M.find s reserved_words
     with Not_found -> SYMBOL s
 
+  let quoted_symbol newline lexbuf s =
+    (* register the newlines in quoted symbols to maintain correct locations.*)
```
bclement-ocp, line 142: But of course (:

bclement-ocp (approved):
Sorry, I linked this PR somewhere but still forgot to look at it.

I think this fits what we need on the Alt-Ergo side :)
It is unfortunate that the `Print` modules for smt and psmt can't be shared…

src/languages/smtlib2/poly/lexer.mll
Gbury: of course, ^^
note that the newline accounting does two things: 1) update the lexer internal counters for positions and 2) update the `dolmen`-specific table of line starts for a file (which is used to store compact locations, and later recover line counts if needed)

PR #199, opened 2023-11-13
Add proper support for `:named` smtlib annotations

Gbury:
The `:named` smtlib annotation has relatively annoying semantics. In particular, it implicitly introduces a global definition. Therefore this PR does a few things:
- Add a generic support for implicit definition/declarations that can ben generated during typechecking, and propagate that change. In particular this means that the result of typechecking one "thing"( be it a term/formula/declaration/definition) is actually a list of implicit declarations, a list of implicit definitions, and then the result of typechecking that "thing". Also, typechecking a single statement now results in multiple statements. This PR decides to keep these as a list, instead of exploding them as could be possible with an adequate `Pipeline` transformation. This is because if we exploded them, then the intermediate typechecking state could be surprising to downstream pipes, since it would contain statements that these pipes have not yet seen.
- That support is used by the typechecker to emit the various implicit definition and declarations that were already recorded but not emitted, for instance, all the constants (but not the variables) that are inferred.
- A new `Hook` result for tags is added in the typechecker to allow for tags to arbitrarily change the typechecked term (instead of simply adding/setting some flags/tags in terms)
- Lastly, the `core` theory of smtlib2 is adapted to use the `Hook` mechanism to properly replace named expression by their symbol, which is required since named expressions can be referred later by their symbol.


doc/tuto.md
```diff
@@ -69,8 +69,8 @@ let test file =
   (* We can loop over the parsed statements to generated the typed statements *)
   let final_state, rev_typed_stmts =
     List.fold_left (fun (state, acc) parsed_stmt ->
-      let state, typed_stmt = Typer.check state parsed_stmt in
-      (state, typed_stmt :: acc)
+      let state, typed_stmts = Typer.check state parsed_stmt in
+      (state, List.rev_append typed_stmts acc)
```
bclement-ocp, line 73: Is this `rev_append` yielding the correct final order? In what order should we process the `typed_stmts` that come out of `Typer.check`?

src/loop/typer_intf.ml
```diff
@@ -268,11 +279,11 @@ module type S = sig
   val print : Format.formatter -> typechecked stmt -> unit
   (** Printing funciton for typechecked statements. *)
 
-  val check : state -> Dolmen.Std.Statement.t -> state * typechecked stmt
+  val check : state -> Dolmen.Std.Statement.t -> state * typechecked stmt list
```
bclement-ocp, line 282: Maybe would be worth documenting why this returns a list and its order? I think implicits are first?

src/typecheck/core.ml
```diff
@@ -598,10 +594,43 @@ module Smtlib2 = struct
...
+                       let f = Type.T.Const.mk path ty in
+                       (* Ids in attributes are created with namespace Attr, but named ids
+                          are meant to be also used in terms, so we need to tweak the
+                          namespace before declaring the binding. *)
+                       let bound_id = { id with ns = Term } in
+                       Type.decl_term_const env (Type.Ast res_ast) bound_id f
+                         (Type.Implicit_in_term (Type.file env, res_ast));
+                       Type.register_implicit env
+                         (`Term_def (bound_id, f, [], [], t));
+                       Type.T.Const.set_tag f Tag.named "";
```
bclement-ocp, line 626: It probably doesn't change much but I'd call this immediately after `let f = Type.T.Const.mk path ty in` (just so that the non-tagged version doesn't leak if `decl_term_const` or `register_implicit` ever throw).

bclement-ocp (approved):
That looks good to me, I like the design of the `Hook` capability. Just a few minor questions and comments.

doc/tuto.md
Gbury: It should yield the correct order **ultimately**, indeed note that this fold produces the list of statements **in reverse order**, which is underscored by the fact that the result is bound to a variable named `rev_typed_stmts`.

The statements returned by `check` should be treated in the natural order of the list (one should be able to print the statements in the order provided, and get a problem equivalent to the one from the input).

src/loop/typer_intf.ml
Gbury: doc added

src/typecheck/core.ml
Gbury: done

src/loop/typer.ml
```diff
@@ -1784,53 +1832,96 @@ module Typer(State : State.S) = struct
...
+
+  (* definitions helpers *)
+  let tr_def ~implicit ~recursive typed_def =
+    match typed_def with
+    | `Type_alias (id, c, vars, body) ->
+      if not recursive && not implicit then begin
+        Dolmen.Std.Expr.Ty.alias_to c vars body;
+        `Type_alias (id, c, vars, body)
+      end else
+        assert false (* TODO: proper exception *)
```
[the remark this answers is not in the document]
Gbury: fixed in the last commit

Issue #200, opened 2023-11-17
Support non-incremental parsing from stdin


bclement-ocp:
Some parsers (e.g. Alt-Ergo's native format) do not support incremental parsing, because there is no clear delimiter between phrases. 
Using such formats with an `` `Stdin`` input throws an exception, when it should just switch to the full parser.

This would obviously not allow interactive input for these parsers, but would still allow piping files in this format through stdin.

See also OCamlPro/alt-ergo#949

PR #201, opened 2023-11-17
Add full mode stdin parsing

Gbury:
Also, take the opportunity to clean up the logic class interface. Should solve #200 

Note that even with this patch, in order to parse stdin using alt-ergo's native format, one need to use the new `parse_all` function, because the `parse_input` function will still raise. That being said the parsing pipe (in the `loop` sub-library) will adequately switch modes as needed and should allow full stdin parsing for all languages (as long as there is an end of input from stdin).

cc @bclement-ocp

### DIFF
--- a/src/classes/logic.ml
+++ b/src/classes/logic.ml
@@ -21,9 +21,6 @@ module type S = sig
   val find :
     ?language:language ->
     ?dir:string -> string -> string option
-  val parse_file :
-    ?language:language ->
-    string -> language * file * statement list
   val parse_all :
     ?language:language ->
     [< `File of string | `Stdin of language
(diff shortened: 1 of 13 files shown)


src/classes/logic.ml
```diff
@@ -174,6 +164,28 @@ module Make
...
+      let locfile, res = P.parse_file_lazy file in
+      l, locfile, res
+    | `Raw (filename, l, s) ->
+      let language =
+        match language with Some _ -> language | None -> Some l
+      in
+      parse_raw_lazy ?language ~filename s
+    | `Stdin l ->
+      let filename = "<stdin>" in
+      let s = Dolmen_std.Misc.read_all ~size:1024 stdin in
```
bclement-ocp, line 183: This works — it is what we are currently doing in Alt-Ergo — and is probably fine, but I was thinking doing this inside Dolmen would make it possible to create the `lexbuf` with `Lexing.from_channel stdin` directly, avoiding an extra copy (i.e. I was thinking of an equivalent of `parse_raw_lazy` that takes an `in_channel` rather than a string buffer — I can make that patch if you'd like).

PR #204, opened 2024-01-26
[WIP] Add Tptp v8.2

Gbury:
This add TPTP v8.2 to Dolmen.

This is still a work in progress, and there are quite a few things to add (mainly around typechecking).


Gbury:
This is outdated, since TPTP `v9.1.0` has been recently released, I'll see when i have time to update the TPTP parser.

Issue #205, opened 2024-02-27
Question regarding BV logic in SMT files


rod-chapman:
I have many SMT2.6 files, all of which use the "BV" logic for BitVectors, and (in particular) use the "bv2nat" function that is supported by Z3, CVC4 and CVC5.

Dolmen appears to reject this - e.g.
```
204 |   (ite (bvsge x (_ bv0 16)) (bv2nat x) (- (- 65536 (bv2nat x)))))
                                   ^^^^^^
Error Unbound identifier: `bv2nat`
```

Any workaround for this?

Gbury:
Unfortunately, there is no current workaround, but I'm planning to work on one in the coming days.

For a bit of context: although supported by Z3, CVC4 and CVC5, `bv2nat` is not part of the official standard defined by the smtlib. More specifically, [the theory](https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml) for BV internally defines `bv2nat` but only so that it can be used in its semantic specification.

Until now, `dolmen` has strived to implement the standard, at least as a first goal, so that it could be used to check conformity against that standard. However, it appears that it's not sufficient anymore, so I'm planning to add a notion of typing extensions: these extensions would be enabled by cli flags and effectively add new functions to the typing builtins (similar to how the parsing extensions work, see #190 ).

rod-chapman:
Understood. I'm trying to get a non-trivial program verification benchmark into the SMTComp suite. The code is cryptographic, so it has modular arithmetic and BV stuff all over the place...

Gbury:
I should have an experimental version of `dolmen` that can accept `bv2nat` by the end of the week.

Gbury:
Just to make sure, what would be the expected signature of the `bv2nat` function ?

rod-chapman:
Not sure.... how is it declared in Z3, CVC4 and CVC5?

bclement-ocp:
As far as I could tell when implementing these functions for Alt-Ergo, they use a family of function symbols of the form `(bv2nat (_ BitVec n) Int)` for all `n > 0`, and indexed function symbols `((_ int2bv n) Int (_ BitVec n))` for `n > 0`.

(I believe Z3 also supports `nat2bv` but CVC5 only supports `int2bv` because the developers didn't like `nat2bv` allowing negative arguments)

Gbury:
#208 adds a way for `dolmen` to accept the `bv2nat` and `int2bv` functions.

[...]

PR #208, opened 2024-03-01
Add typing extensions, and the bvconv extension

Gbury:
Should solve #205 

This PR adds a notion of typing extensions, which allow to add new builtins for the typechecking. Fomr the point of view of the ocaml API, these do not add much compared to the already existing `additional_builtins` mechanism. They mainly represent a slightly more structured way to add new builtins. When using the `dolmen` binary, these extensions can be activated on the command line, using the new `--ext` option.

Currently, these is only a single extension: the `bvconv` extension, which adds the `bv2nat` and `int2bv` functions, as described in https://github.com/Gbury/dolmen/issues/205#issuecomment-1969393448

Therefore, examples of smtlib2 problems using either `bv2nat` or `int2bv` will now be able to typecheck using the following command-line:
```
dolmen --ext=bvconv problem.smt2
```


rod-chapman:
Thanks. Let me know when it's merged.

PR #214, opened 2024-05-06
feat: Add support for user-defined builtins with Dune plugins

### LINKED ISSUE #203: Supporting evaluation for custom functions
If an input file contains custom built-in functions, Dolmen's binary cannot check the model as Dolmen's evaluation function doesn't know this built-in. 

For instance `Alt-Ergo` has a custom built-in `round` function which isn't a part of the SMT-LIB standard. 
As we discussed offline, this low-priority feature is annoying to implement.

A naive way to support this feature in AE without modifying Dolmen consists in adding definitions for these builtins in the input problem which means we have to add an option in AE to check models with Dolmen's library instead of calling the binary on the outputted model by the solver.

bclement-ocp:
This patch adds support for model extensions in `Dolmen_loop` using a similar mechanism to typing extensions.

Model extensions and typing extensions are now loaded by the binary through Dune's plugin mechanism, and the existing `bvconv` extension is moved to use the plugin mechanism.

Fixes #203


bclement-ocp:
> the existing `bvconv` extension is moved to use the plugin mechanism

In retrospect, I am not 100% sure this is a good idea. I did this initially to have an example of usage of the plugin mechanism, but ended up writing two example plugins (`abs_real`, `abs_real_split`) which I think is better for that purpose since. I am open to reverting this part of the change.

bclement-ocp:
> > the existing `bvconv` extension is moved to use the plugin mechanism

This was causing build failures on the CI so no longer part of the PR. Instead there is support for builtin extensions when loading them at the CLI level.

bclement-ocp:
I did some late cleanup but this should be ready to review @Gbury

src/bin/options.ml
```diff
@@ -268,16 +269,183 @@ let report_style =
...
+
+let extension =
+  let parse s =
+    match parse_ext_opt s with
+    | None -> Fmt.error_msg "Invalid extension name '%s'" s
+    | Some (name, kind) ->
+      Result.map (fun ext -> (ext, kind)) (find_ext name)
+  in
+  let print ppf _p = Fmt.pf ppf "" in
+  Arg.conv (parse, print)
```
Gbury, line 448: I think this is enough code to warrant it being mostly in a new file (`extensions.ml` ?).

src/bin/options.ml
```diff
@@ -268,16 +269,183 @@ let report_style =
...
+        | Some _ -> Unavailable
+      in
+      { extension_name ; typing_plugin ; model_plugin } :: ok, err
+  in
+  let ok, err =
+    List.fold_left (fun (ok, err) e ->
+      add_plugin ok err e @@ parse_ext_opt e
+    ) (builtin_extensions, []) (Dolmen.Sites.Plugins.Plugins.list ())
+  in
+  merge_extensions ok, List.fast_sort String.compare err
```
Gbury, line 385: Hm... I'm a bit worried by the cost of computing the list of all extensions systematically at startup (particularly since listing all dune plugins might involve a number of system calls).
Intuitively, we should only need that we want want to list all extensions, i.e. when a user asks for a list of extensions (or in error messages), but when loading an extension, we should be able to not need the lis tof all extensions.

src/loop/typer.ml
```diff
@@ -73,6 +73,35 @@ module Zf_arith =
...
+(* ************************************************************************ *)
+
+module Ext = struct
+
+  type t = {
+    name : string;
+    builtins : Typer_intf.lang -> T.builtin_symbols;
+  }
+
+  let all = ref []
```
Gbury, line 86: Considering that one of the main use of extensions (in the cli) is loading them using their name, maybe a hashtable indexed by names would be better instead of a list here ? That would avoid surprising behaviour later if/when we start having a fair number of extensions (unlikely, but we never know).

Gbury (commented):
Overall this looks good.

Note: you can rebase on current master to fix the CI bugs.

src/bin/options.ml
bclement-ocp: I think I did it this way because I initially wanted to preserve the behavior where the list of all available extensions is provided as an `enum` to Cmdliner. Since this is no longer the case (and there is `--list-extensions` instead), it makes sense to perform the listing lazily; will change.

src/bin/options.ml
bclement-ocp: Fair enough, will do.

src/loop/typer.ml
bclement-ocp: I simply reused the previous implementation here. I agree it would be better in the unlikely case we have a lot of extensions to use a hash table; I'd prefer changing the implementation in a follow-up PR if that's OK.

Gbury: that's ok

bclement-ocp:
Made the requested changes + some others following (old) offline discussions:

 - Move code dealing with extensions out of the Options module and into
   a new Extensions module

 - Clean up examples (notably avoid an almost-empty plugin in
   abs_real_split by defining the builtin in the typing plugin; it does
   not make sense to use the model plugin without the typing plugin)

 - Load dune plugins lazily

 - Use a hash table to find binary extensions (but not typing
   extensions/model extensions)

 - Allow override of builtin extensions by dune plugins

src/loop/typer.ml
bclement-ocp: Ended up actually doing it in this PR

bclement-ocp:
@Gbury ping

bclement-ocp:
It's been a while, I don't remember the conclusion of our last discussion. I'll go over the code again while applying review changes (probably sometime next week); if you are happy with the current architecture I'd say we keep it the way it is but if there are architectural issues we can fix them.

dolmen_bin.opam
```diff
@@ -15,6 +15,7 @@ depends: [
   "dolmen_loop" {= version }
   "dolmen_model" {= version }
   "dune" { >= "3.0" }
+  "dune-site" { >= "3.0" }
```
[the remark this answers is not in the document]
bclement-ocp: The `--relocatable --prefix $dir` options are if one wants a relocatable installation (i.e. with paths to plugins relative to the executable). I think it's safe to treat this as a packaging issue and not worry about it on our side.

dune-project
```diff
@@ -2,4 +2,24 @@
 (name dolmen)
 (using mdx 0.2)
 (using menhir 2.0)
+(using dune_site 0.1)
 (implicit_transitive_deps true)
+
+(package
+  (name dolmen)
+  (sites (lib plugins)))
```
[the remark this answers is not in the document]
bclement-ocp: I followed dune-site's recommendation:

> Traditionally, this is in (lib plugins), but it’s just a convention.

More precisely my understanding is that the plugins site contains platform-dependent code (specifically dynamic libraries) so it makes sense to be in `lib` (whereas `share` is usually for platform-independent data).

bclement-ocp:
Rebased, looking at the review now.

src/bin/extensions.ml
```diff
@@ -0,0 +1,235 @@
...
+  | Unavailable, Unavailable -> Unavailable
+
+type t = {
+  extension_name : string ;
+  (** Name of the extension (without dot). *)
+  typing_plugin : Dolmen_loop.Typer.Ext.t location ;
+  (** Name of the Dune plugin that provide typing extensions. *)
+  model_plugin : Dolmen_model.Ext.t location ;
+  (** Name of the Dune plugin that provide model extensions. *)
+}
```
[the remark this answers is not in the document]
bclement-ocp: I usually try to define functions that operate on a type close to the type definition, but defining all types first works as well. Done.

src/bin/extensions.ml
```diff
@@ -0,0 +1,235 @@
...
+    (** An external extension to be loaded dune's plugin mechanism. Loading the
+        plugin should register an extension in the appropriate registry. *)
+(** The ['a location] type represents the location of an extension (either
+    built-in, or to be loaded from a specific dune plugin). *)
+
+let is_available = function
+  | Unavailable -> false
+  | Builtin _ | Dune_plugin _ -> true
+
+(* Merge two possible locations for the same plugin. Prefer builtin plugins. *)
```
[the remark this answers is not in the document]
bclement-ocp: I think the use case was to be able to fix or extend an existing plugin. But it's probably too brittle; I'll make it an error.

examples/extensions/abs_real/dune
```diff
@@ -0,0 +1,10 @@
+(library
+  (public_name dolmen-extension-abs-real)
+  (name abs_real)
+  (modules abs_real)
+  (libraries dolmen_model))
```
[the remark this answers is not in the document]
bclement-ocp: Yup, it does — fixed.

bclement-ocp:
Updated code for loading builtin extensions. I'm a bit unhappy about the fact that other modules loaded by the binary could create new extensions that would not get picked up (depending on module load order), but I think that's mostly a theoretical concern --- builtin extensions should be defined in the shared extension module, and plugin extensions should be defined in plugins (which are loaded dynamically and won't be seen when listing extensions at toplevel).

src/bin/extensions.ml
```diff
@@ -0,0 +1,139 @@
+(* This file is free software, part of dolmen. See file "LICENSE" for more information *)
+
+type location =
+  | Builtin
+    (** A built-in extension. *)
+  | Dune_plugin of string
+    (** An external extension to be loaded using dune's plugin mechanism.
+        Loading the plugin should register an extension in the appropriate
+        registry. *)
+(** The ['a location] type represents the location of an extension (either
```
Gbury, line 10:
```suggestion
(** The [location] type represents the location of an extension (either
```

src/bin/extensions.ml
```diff
@@ -0,0 +1,139 @@
...
+    Hashtbl.replace builtin_extensions (Dolmen_loop.Typer.Ext.name ext) ()
+  );
+
+  (* Add builtin model extensions (note: these may share name with typing
+     extensions). *)
+  Dolmen_model.Ext.iter (fun ext ->
+    Hashtbl.replace builtin_extensions (Dolmen_model.Ext.name ext) ()
+  );
+
+  builtin_extensions
```
Gbury, line 45: there are quite a few blank lines that I think could be removed here ?

src/bin/extensions.ml
```diff
@@ -0,0 +1,139 @@
...
+  lazy (
+    let extensions = Hashtbl.create 17 in
+
+    List.iter (fun plugin ->
+      if Hashtbl.mem builtin_extensions plugin then
+        Format.eprintf
+          "@[<v>Warning: @[<hov>%a@ '%s'@ %a@]@]@."
+          Fmt.text "Ignoring external extension"
+          plugin
+          Fmt.text "as it would override a builtin extension of the same name."
```
Gbury, line 57: Instead of a lazy, could we have this be a function that takes a `State.t` and can therefore use all of the nice reporting facilities for warnings, errors and so on ?

src/bin/extensions.ml
```diff
@@ -0,0 +1,139 @@
...
+  | Dune_plugin plugin ->
+    if Hashtbl.mem loaded_plugins plugin then Ok ()
+    else
+      match Dolmen.Sites.Plugins.Plugins.load plugin with
+      | () ->
+        Hashtbl.replace loaded_plugins plugin ();
+        Ok ()
+      | exception Dynlink.Error err ->
+        Fmt.error_msg "while loading plugin %s: %s"
+          plugin (Dynlink.error_message err)
```
Gbury, line 99: Same remark as above, it would be nice if this could use the generic warning/error reporting facilities

src/bin/extensions.ml
```diff
@@ -0,0 +1,139 @@
...
+  | [] -> Ok st
+  | exts  -> Ok (add_model_extensions exts st)
+
+let parse extension_name =
+  let extension_location =
+    if Hashtbl.mem builtin_extensions extension_name
+    then Ok Builtin
+    else
+      (* Dune doesn't provide a way to check if a plugin with a given name is
+         available, so list all plugins to check. *)
```
Gbury, line 129: instead of comparing the plugin name with the list of everything available, wouldn't it be possible to catch some kind of errors that I expect to be raised when trying to load the plugin ?

src/bin/options.ml
```diff
@@ -267,18 +268,11 @@ let report_style =
...
-let typing_extension_list =
-  List.map
-    (fun ext -> Loop.Typer.Ext.name ext, ext)
-    (Loop.Typer.Ext.list ())
-
-let typing_ext =
-  Arg.enum typing_extension_list
-
+let extension =
+  Cmdliner.Arg.conv (Extensions.parse, Extensions.pp)
```
Gbury, line 275: As stated in other part of this review, I think it'd be better to extensions just be strings when parsing them as option values, and then later, once we have a `State.t`, we can load them, and if any problem arises, report it using the warning/error mechanism.

Gbury: In terms of code architecture, that would also help in not needing a lazy value for the list of plugins, which can then be evaluated once (just before loading all plugins) without the need for a lazy.

src/bin/options.ml
```diff
@@ -731,4 +741,11 @@ let cli =
     let doc = "The warning or error of which to show the documentation." in
     Arg.(value & opt (some mnemonic_conv) None & info ["doc"] ~doc ~docv:"mnemonic" ~docs)
   in
-  Term.(ret (const aux $ state $ preludes $ logic_file $ list $ doc))
+  let list_extensions =
+    let doc = "List all available extensions." in
+    Arg.(value & flag & info ["list-extensions"] ~doc ~docs)
```
Gbury, line 746: maybe we could use `--ext-list` to avoid sharing the same prefix as `--list` ? (although admittedly using `--list` for report lists wasn't a great idée on my part).

Gbury (changes requested):
Thanks a lot ! I have a few comments on how warnings and errors are reported around plugin loading, and after that it'll be good to merge.

src/loop/typer_intf.ml
```diff
@@ -1,4 +1,9 @@
-(* This file is free software, part of Archsat. See file "LICENSE" for more details. *)
+(* This file is free software, part of dolmen. See file "LICENSE" for more details. *)
+
+type lang = [
+  | `Logic of Logic.language
+  | `Response of Response.language
+]
```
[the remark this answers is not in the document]
bclement-ocp: I just moved this code to make the `lang` type available from outside the functor, what would you want the comment to say?

src/bin/extensions.ml
bclement-ocp: I think the blank lines make it more readable, I can remove them.

src/bin/extensions.ml
bclement-ocp: I don't think this works if we want to be able to list the plugins when using `--list-extensions`.

src/bin/extensions.ml
bclement-ocp: Fair, that one is only called once we have a state.

src/bin/extensions.ml
bclement-ocp: The error that is raised is unfortunately private and can't be caught without a catch-all `try .. with` clause.

src/bin/options.ml
bclement-ocp: There is a tension between using Dolmen's error mechanism and using Cmdliner's error mechanism. I think for parsing command line options using Cmdliner makes more sense, but I can switch to using Dolmen's errors instead.

It's not entirely clear to me what that would entail in the `--list-extensions` case however.

src/bin/options.ml
bclement-ocp: Well then we share a prefix with `--ext` (: I'm find with changing the flag though.

src/loop/typer.ml
```diff
@@ -1122,7 +1127,8 @@ module Typer(State : State.S) = struct
     State.create_key ~pipe "smtlib2_forced_logic"
   let extension_builtins : Ext.t list State.key =
     State.create_key ~pipe "extensions_builtins"
-  let additional_builtins : (state -> lang -> T.builtin_symbols) State.key =
+  let additional_builtins :
+    (state -> Typer_intf.lang -> T.builtin_symbols) State.key =
```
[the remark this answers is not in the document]
bclement-ocp: I don't think so, but I didn't want to break the existing API.

src/loop/typer.mli
```diff
@@ -40,6 +40,38 @@ val bad_arith_expr : (Dolmen_type.Arith.Smtlib2.config * string) Report.Warning.
 val unknown_logic : string Report.Warning.t
 (** Unknown logic warning *)
 
+module Ext : sig
+  (** Define typing extensions.
+
+      These extensions are typically extensions used by some community,
+      but not yet part of the standard.
+
+      @since 0.10 *)
```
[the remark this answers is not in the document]
bclement-ocp: The code was written before 0.10, good catch!

src/bin/options.ml
Gbury: I really prefer dolmen's mechanism, particularly for anything like a warning, since with dolmen's mechanism we can make the warning fatal if wanted.

I'm not sure what problem you see with the  `--list-extensions` case ? Basically, i'm arguing that we should do less (actually the minimal amount of) work when parsing the options, and more work afterwards, once we have the state. In the case of extensions, that means not using cmdliner to check whether the extension exists when parsing (for the parsing extensions are simple arbitrary strings, no real parsing) and instead wait until we try and load the extension to error out.

src/bin/extensions.ml
Gbury: I'm fine with using a catch_all around the loading of the plugin, since in case of an exception, we will only print things, and then exit, which is reasonable for basically any exception I can think of ? Something along the lines of:
```ocaml
match Dolmen.Sites.Plugins.Plugins.load plugin with
| () -> ...
| exception exn -> (* report the error *)
```

src/bin/options.ml
bclement-ocp: We don't have a Dolmen state for ``--list-extensions` is the problem I see.

src/loop/typer_intf.ml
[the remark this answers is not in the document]
Gbury: Just something like `(** Langues for files parsed (and typed) by dolmen *)` ?

src/loop/typer.ml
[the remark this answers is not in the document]
Gbury: Agreed, We should have at least one version with both mechanisms available so that users can switch more easily (I have created #232 to not forget about it).

src/bin/extensions.ml
bclement-ocp: That's fair actually, I can do this.

src/bin/extensions.ml
Gbury: I don't understand why that would be a problem: what prevents us from creating a state before computing the list of extensions in the `--list-extensions` case ?

src/bin/options.ml
Gbury: Ok, but we can create one, can't we ? Put another way: is there any missing information that would prevent us from creating a state for that case ?

src/bin/extensions.ml
bclement-ocp: Actually you are right, we do have a `State.t` which we throw away.

bclement-ocp:
Dropped the lazy and moved to use Dolmen's warning and errors.

src/bin/extensions.ml
bclement-ocp: I ended up getting rid of this warning, I don't think it's useful to have it here. But I can move it to the actual loading time actually, let me do that.

Gbury:
There's a CI failure: very likely, we need an additional lower bound on `dune` to use `sites`, and this will also induce a lower bound on the ocaml compiler version I suppose.

src/bin/extensions.ml
bclement-ocp: Actually I can't because I can't easily check if a plugin exists as a dune plugin or not…

bclement-ocp:
I have moved the `dune-site` dependency to the binary only since only the binary needs the `Sites` module now. Hopefully that should work.

bclement-ocp:
Ah, this is actually a small issue — we need to decide whether we want plugins to be 100% a dolmen binary thing (and put them in the `dolmen_bin` namespace) or if we want them to go into the `dolmen` namespace (so that installed plugins could be accessed by other users of the Dolmen library).

I think both choices make sense — using the `dolmen_bin` namespace is a bit cleaner in terms of dependencies, but using the `dolmen` namespace would allow seamless integration with other tools using the Dolmen library (e.g. alt-ergo could pick up installed Dolmen plugins, which is nice). I'll let you make the call (in both cases this does not require changing the lower bound on either `dune` or the compiler -- just adding a dependency to `dune-site >= 3.0` in addition to `dune >= 3.0` for the appropriate package).

src/dolmen.ml
```diff
@@ -15,3 +15,5 @@ module Zf = Dolmen_zf
 (* Classes *)
 module Class = Dolmen_class
 
+(* Extensions *)
+module Sites = Sites
```
Gbury, line 19: I'm wondering: since we expect users of the `dolmen` lib to (potentially) use extensions/plugins, should be directly expose this module, or should we expose a thin wrapper around it ? I'm worried that `dune` may one day change the interface, and lead to weird compilation errors, and that the module may lack some documentation ?

src/bin/extensions.ml
```diff
@@ -0,0 +1,125 @@
...
+  let all_extensions =
+    Hashtbl.fold
+      (fun extension_name () exts ->
+        extension_name :: exts)
+      builtin_extensions
+      []
+  in
+  let all_extensions =
+    List.fold_left (fun all_extensions plugin ->
+      if Hashtbl.mem builtin_extensions plugin
```
Gbury, line 41: In that case, we could emit a warning to warn that an installed plugin with the same name as a builtin plugin will never be usable, as the builtin will always be loaded instead ? Even if it is only triggered when explicitly asking for the list of all available plugins with the command line, I think that would be useful.

src/bin/extensions.ml
```diff
@@ -0,0 +1,125 @@
...
+      Hashtbl.replace loaded_plugins plugin ();
+      st
+    )
+    else
+      match Dolmen.Sites.Plugins.Plugins.load plugin with
+      | () ->
+        Hashtbl.replace loaded_plugins plugin ();
+        st
+      | exception Dynlink.Error err ->
+        Loop.State.warn st dynlink_error (plugin, err)
```
Gbury, line 95: In that case, are we sure that the plugin has been actually loaded ? Else this could probably be an error ?

src/bin/extensions.mli
```diff
@@ -0,0 +1,29 @@
...
+
+val pp : Format.formatter -> t -> unit
+
+val name : t -> string
+(** Returns the name of the extension. *)
+
+val load_typing_extension :
+  t -> Loop.State.t -> (Loop.State.t, [> `Msg of string]) result
+(** [load_typing_extension e] loads and returns the typing extension associated
+    with [e].
```
Gbury, line 11: This does not return the typing extension, but rather the state after loading the extension.

src/bin/extensions.mli
```diff
@@ -0,0 +1,29 @@
...
+
+    Fails if an error occurs during loading of
+    an external typing extension. *)
+
+val load_model_extension :
+  t -> Loop.State.t -> (Loop.State.t, [> `Msg of string]) result
+(** [load_model_extension e] loads and returns the model extension associated
+    with [e].
```
Gbury, line 19: Same remark as above.

src/bin/extensions.mli
```diff
@@ -0,0 +1,29 @@
...
+
+    Fails if [e] has no model extension, or an error occurs during loading of an
```
Gbury, line 21: the function does not fail if [e] has no model extension

src/bin/options.ml
```diff
@@ -384,26 +378,39 @@ let mk_run_state
...
+    |> Loop.Flow.init ~flow_check
+    |> Loop.Header.init
+      ~header_check
+      ~header_licenses
+      ~header_lang_version
+  in
+  (* Extensions *)
+  let st =
+    List.fold_left (fun st ext ->
+      Result.bind st (Extensions.load_typing_extension ext)
```
Gbury, line 405: the `Extensions.load_[typing|model]_extension` functions now always return `Ok`, so do we really want to keep the `Result.bind` style of code here ? Particularly since we will use dolmen's warning/error mechanism to report problems during that code.

Gbury (commented):
Last round of questions !

src/bin/extensions.ml
bclement-ocp: I think we are sure that the plugin has *not* been loaded. It's probably better to always fail if the plugin is not loaded though, rather than let execution continue when a plugin that was explicitly requested by the user failed to load. Will change to an error.

src/bin/options.ml
bclement-ocp: No, this (and above comments) is because I didn't finish the refactoring I suppose. Will fix.

src/dolmen.ml
bclement-ocp: I think that the `dune-site` interface is fairly stable. We could wrap it / add a module signature, but then if `dune-site` *does* add new features they won't be available until we change the wrapper / module signature, so I don't know how much I like this idea.

I can add some doc that this is a `dune-site` module and should only be used by expert users though.

Gbury: That works.

src/bin/main.ml
```diff
@@ -121,11 +121,27 @@ let doc conf t =
...
+        and will never be loaded"
+        Fmt.(hovbox @@ list ~sep:Fmt.comma (box string)) conflicts
+    )
+    ~name:"Conflicting extensions" ()
+
+let list_extensions st =
+  begin match Extensions.list_conflicts () with
+  | [] -> ()
+  | conflicts ->
+    ignore (Loop.State.warn st conflicting_extensions conflicts)
```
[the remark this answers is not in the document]
Gbury: ah nevermind, I read it wrong.

Gbury (approved):
Great, this is good to merge ! If you want to write the `CHANGES.md` entry you can, else I'll do it right after merging.

bclement-ocp:
Done!

Gbury (approved):

PR #220, opened 2024-09-06
feat: Support more Alt-Ergo primitives

bclement-ocp:
This patch adds typing support for the following Alt-Ergo primitives:

 - `fpa_rounding_mode` and its constructors (equivalent to `RoundingMode` in SMT-LIB);
 - `real_of_int` (equivalent to `to_real` in SMT-LIB);
 - `real_is_int` (equivalent to `is_int` in SMT-LIB);
 - `abs_int` (equivalent to `abs` in SMT-LIB);
 - `int_floor` (equialent to `floor_to_int` in SMT-LIB)

**Note**: These have been selected because they map directly to SMT-LIB types and primitives. There are additional floating-point (and real) primitives that do not exist at the SMT-LIB level, such as the `round` function. It is not clear at this point whether they should be added in Dolmen itself or in a separate plugin.

### DIFF
--- a/src/interface/term.ml
+++ b/src/interface/term.ml
@@ -761,6 +761,9 @@ module type Ae_Arith = sig
     val to_real : t -> t
     (** Conversion from an integer term to a real term. *)
 
+    val abs : t -> t
+    (** Arithmetic absolute value. *)
+
   end
 
   module Real : sig
@@ -768,6 +771,12 @@ module type Ae_Arith = sig
 
     val div : t -> t -> t
     (** Exact division on reals. *)
+
+    val is_int : t -> t
+    (** Integer testing *)
+
+    val floor_to_int : t -> t
+    (** Greatest integer smaller than the given real *)
 
   end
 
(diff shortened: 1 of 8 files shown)


bclement-ocp:
After thinking about it I am a bit on the fence on the topic. I would like the Dolmen experience to be consistent w.r.t builtins: either a builtin is fully supported by dolmen, including printing to SMT-LIB, or it is not at all and requires a plugin (once the corresponding PR is merged). Unfortunately this is already not the case (due to e.g. exponentiation and `abs_real`).

I'll think about it for a bit; in the meantime, I think this PR is uncontroversial and can be merged.

src/interface/term.ml
```diff
@@ -931,6 +940,46 @@ module type Ae_Bitv = sig
 
 end
 
+module type Ae_Float_Float = sig
```
Gbury, line 943: Is this module type necessary ? Why not just manually inline it at its only use site ? Or do you envision it being used elsewhere ?

Gbury (approved):
Just one small question, but otherwise good to merge

PR #221, opened 2024-09-12
fix: Add missing model for `to_int` coercion

bclement-ocp:


Gbury:
Thanks for that fix to some obvious missing code, ^^

PR #235, opened 2025-04-18
Support for SMT-LIB 2.7

Gbury:
Still WIP, but should be nearly complete.


src/interface/term.ml
```diff
@@ -350,6 +357,13 @@ module type Logic = sig
   (** S-expressions (for smtlib attributes), should probably be related
       to the [data_t] term. *)
 
+  val smt2_clusterfuck : ?loc:location -> t -> t list -> t
+  (** Eldricht horror from SMT-LIB2.7. *)
```
bobot: Naming is hard... ;)

Gbury: In the case of smt2.7, the higher-order encoding will be represented as such in dolmen terms (i.e. using dedicated builtins and binders), so there is a (small) cost to using the higher-order encodings (compared to the first order terms). So I'd recommend only using higher-order when necessary, **and** being explicit in the printers, i.e. print first order applications using the regular syntax, and always print higher order application using the `@` symbol).

bobot: Of course, only partial application, and anonymous function (lambda term) would use the new syntax. The question was more to use 2.7 instead of 2.6+par for Alt-Ergo, COLIBRI and colibri2.

Gbury: Hm... personally I prefer 2.6+par, but the SMT-LIB maintainers seem to want to stuck with their `declare-sort-param` which seems likely to stay for smtlib3, so it might make sense to try and go towards it at one point...
In any case, both will be supported by dolmen, the semantics are comparable, and I don't expect there to be a significant difference in performance (though the 2.7 version will be slightly slower), so for 2.7 it shouldn't matter much.

Gbury:
With the last commit, all of the new features of SMT-LIB2.7 have been implemented[1]. Once tests pass, I'll probably merge soon.

@bobot depending on how far you are with the why3 printers, would you happen to have a few files that make use of the new features (e.g. polymorphism, higher-order, check-sat-assuming with arbitrary terms) ? That would be helpful to check that the implementation is correct and works as intended.

[1] : though I still have to try and find a sane way to restrict the new functions of the `BV` theory  (i.e. `ubv_to_int`, `sbv_to_int`, `int_to_bv`) to only be present if the `Int` theory is also present...

PR #240, opened 2025-05-22
Hierarchical builtins

Gbury:
This PR adds some hierarchy in the typed builtins. Previously all the builtins lived in a flat namespace, but now there is (most of the time) one extensible constructor builtin by theory, which wraps a dedicated (regular) type for the builtins of that theory. This brings a few benefits:
- We can have an exhaustivity check for functions operating on the primitives of one theory
- we can have more freedom in the naming of builtins, since we have different namespaces for the builtins of each theory
- This may result in a slight speedup, since matching on extensible variants is not very optimized

Note that this is a rather large breaking change, since it will likely require to update any pattern matching on builtins, but the translation should be rather mechanical. It is not completely mechanical because I took the opportunity to rename some builtins, taking advantage of the now different namespaces for builtins of different theories (and also to avoid repetitions like `| Builtin.Float Fp_foo -> ...`).

cc @bobot , in particular for the new names of primitives in the Float theory.

### DIFF
--- a/src/standard/builtin.mli
+++ b/src/standard/builtin.mli
@@ -382,156 +382,156 @@ type 'a t += Array of 'a Array.t
 
 module Bitv : sig
   type _ t =
-    | T of int
-    (** [Bitv.T n: ttype]: type constructor for bitvectors of length [n].
+    | T of { n: int; }
+    (** [Bitv.T{n}: ttype]: type constructor for bitvectors of length [n].
         Ensures that [n > 0]. *)
     | Binary_lit of string
     (** [Lit s: Bitv.T]: bitvector litteral. The string [s] should
         be a binary representation of bitvectors using characters
         ['0'], and ['1'] (lsb last) *)
     | To_int of { n : int; signed : bool; }
-    (** [To_int(n,signed): Bitv.T(n) -> Int]:
+    (** [To_int{n; signed; }: Bitv.T{n} -> Int]:
         conversion from bitvectors to signed integers. *)
     | Of_int of { n : int; }
-    (** [Of_int(n): Int -> Bitv.T(n)]:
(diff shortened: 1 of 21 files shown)


src/standard/builtin.mli
```diff
@@ -95,670 +98,712 @@ type _ t +=
...
+    (** [Fp_isPositive(s,e): Fp(s,e) -> Prop.T]: test if it is positive *)
+    | Ieee_format_to_fp of int * int
+    (** [Ieee_format_to_fp(s,e): Bv(s+e) -> Fp(s,e)]: Convert from IEEE interchange format *)
+    | To_fp of int * int * int * int
+    (** [Fp_to_fp(s1,e1,s2,e2): RoundingMode -> Fp(s1,e1) -> Fp(s2,e2)]:
+        Convert from another floating point format *)
+    | Of_real of int * int
+    (** [Real_to_fp(s,e): RoundingMode -> Real -> Fp(s,e)]: Convert from a real *)
+    | Of_sbv of int * int * int
+    (** [Sbv_to_fp(m,s,e): RoundingMode -> Bitv.T(m) -> Fp(s,e)]: Convert from a signed integer *)
```
bobot, line 634: Missed renaming. Inlined record can be also a nice addition, since it is already a breaking change.

Otherwise great, the API is a lot safer now. Perhaps adding in the documentation (can be done later) the corresponding smtlib symbol when it exists can help users.

bobot (changes requested):
Just typos, thanks!

src/standard/builtin.mli
Gbury: The doc comments have been fixed to reflect the new names, and I indeed switched all of the `Bitv` and `Fp` builtins to using inline records, except for the literal ones.

PR #249, opened 2025-09-04
Support SMT-LIB v2.7 bit-vector operators in psmt2

bclement-ocp:
The new bit-vector operators from v2.7 are not available in SMT-LIB 2.6 scripts, which makes sense since Dolmen is intended to be able to check adherence to the specification.

On the other hand, there is no standard specification for the psmt2 format, and while it implements polymorphism differently from SMT-LIB 2.7, there is no reason for psmt2 operators to be restricted to those available in version 2.6 of the standard.

Allow the new bit-vector operators from version 2.7 of the standard when parsing files with psmt2 syntax.

### DIFF
--- a/src/languages/smtlib2/print/print.ml
+++ b/src/languages/smtlib2/print/print.ml
@@ -753,20 +753,20 @@ module Make(Config : Config)(Lexer : Lexer with type token := Config.token) = st
 
           | To_int { n = _; signed = true; } ->
             begin match Config.version with
-              | V2_7 -> simple "sbv_to_int"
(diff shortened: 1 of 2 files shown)


bclement-ocp:
By the way, I'm wondering if having this distinction (SMT-LIB 2.6 vs SMT-LIB 2.7) for the supported operators is actually useful. The bit-vector operators are not (as far as I can tell) mentioned in the actual standard, and only the version 2.7 with the extra operators seem to be available on the smt-lib.org website, e.g. https://smt-lib.org/theories-FixedSizeBitVectors.shtml

PR #250, opened 2025-09-08
Add uutf as a dependency to dolmen

bclement-ocp:
Uutf is used by string_unicode_map in dolmen.std, but it is only marked as a dependency of `dolmen_type`.


Gbury (approved):
Good catch, thanks !

PR #259, opened 2026-03-16
Add Integer exponentiation

Gbury:
Add integer exponentiation as added in SMT-LIB2.7, cf the current version of ints ( https://smt-lib.org/theories-Ints.shtml ), and more particularly:

```
 - ** as the exponentiation function if the second argument is non-negative.
    That is (** m n) is m to the power of n.  In particular, (= (** 0 0) 1) holds.
    If n is strictly smaller than 0, then ** is defined so as to satisfy the formula

    (= (** m n) (div 1 (** m (- n))))

    So the following holds for all integers n < 0:
    - (= (** 0 n) (div 1 0))
    - (= (** m n) (** m (- n))) if (= (abs m) 1)
    - (= (** m n) 0)            if (> (abs m) 1)
```

This also adds the `EIA` and `EIRA` sub-logics (aka logic parts, aka theory+restriction) to the logic parsing, and logic detection for printing/export. More testing will be needed to check (and decide) what happens when a problem using exponentiation is exported back to SMT-LIB2.6.

One main question is whether the SMT-LIB semantics are the same as the current exponentiation operator ? I think it's mainly used by alt-ergo, but maybe also colibri2 ? cc @bobot @bclement-ocp


Gbury:
> The was already a builtin for pow?

Yes there was, though it's probably only used by alt-ergo (either in its native language, or its extensions of smtlib).

bclement-ocp:
I can confirm that the alt-ergo semantics for `pow` is the same as the SMT-LIB one for all non-negative exponents (including zero). Alt-Ergo's `pow` builtin was not defined for negative exponents, but I also think that the SMT-LIB definition makes sense.

Gbury:
Ok, so it's fine for exponentiation in both alt-ergo's native language and SMT-LIB to map to the same builtin ? We don't need to distinguish the two ? (as is done with some other operators, most notably division which can have a few different variations)

bclement-ocp:
Yeah, it's fine to use the same operator in both cases. For division it makes sense because because they correspond to different mathematical operators (depending on rounding), but for exponentiation I don't think it should matter — especially given that Alt-Ergo at the moment does not do any sort of reasoning with exponentiation besides computing it on constant arguments.

